PR #4, opened 2017-11-21
Fix bugs in stored-crypto-key codepaths

chriadam:
And add a unit test for this functionality.

### DIFF
--- a/daemon/SecretsImpl/secretsdatabase.cpp
+++ b/daemon/SecretsImpl/secretsdatabase.cpp
@@ -54,7 +54,7 @@ static const char *createSecretsTable =
         "\n CREATE TABLE Secrets ("
         "   SecretId INTEGER PRIMARY KEY AUTOINCREMENT,"
         "   CollectionName TEXT NOT NULL,"
-        "   SecretName TEXT NOT NULL,"
+        "   HashedSecretName TEXT NOT NULL,"
         "   ApplicationId TEXT NOT NULL,"
         "   UsesDeviceLockKey INTEGER NOT NULL,"
         "   StoragePluginName TEXT NOT NULL,"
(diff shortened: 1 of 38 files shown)


chriadam:
@aleksey-mikhaylov @Denis-Semakin this one fixes bugs in the "encrypt/decrypt with a key reference" codepaths, and adds a unit test.

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -2042,10 +2042,6 @@ Sailfish::Secrets::Daemon::ApiImpl::RequestProcessor::getCollectionSecret(
     } else if (!m_authenticationPlugins.contains(collectionAuthenticationPluginName)) {
         return Sailfish::Secrets::Result(Sailfish::Secrets::Result::InvalidExtensionPluginError,
                                         QString::fromLatin1("No such authentication plugin available: %1").arg(collectionAuthenticationPluginName));
-    } else if (m_authenticationPlugins[collectionAuthenticationPluginName]->authenticationType() == Sailfish::Secrets::AuthenticationPlugin::ApplicationSpecificAuthentication
```
chriadam: This check was premature.  There can be cases where the collection's authentication plugin requires in-app authentication method, but user interaction is not necessary (e.g. if the collection is owned by the application and unlocked).  So I removed this premature check, and only perform this check in the cases where user interaction is deemed required (below).

daemon/SecretsImpl/secretsdatabase.cpp
```diff
@@ -63,7 +63,7 @@ static const char *createSecretsTable =
         "   UnlockSemantic INTEGER NOT NULL,"
         "   CustomLockTimeoutMs INTEGER NOT NULL,"
         "   AccessControlMode INTEGER NOT NULL,"
-        "   FOREIGN KEY (CollectionName) REFERENCES Collections(CollectionName),"
+        "   FOREIGN KEY (CollectionName) REFERENCES Collections(CollectionName) ON DELETE CASCADE,"
```
chriadam: We cannot cascade to the KeyEntries table which is... unfortunate.  I will have to add a trigger to delete from the KeyEntries table, unless someone can think of a better solution..?

chriadam: I guess one alternative is to have "SecretsKeyEntries" for keys which are stored in a Secrets storage plugin, and "CryptoKeyEntries" for keys which are stored by a Crypto plugin directly.  Then the SecretsKeyEntries CollectionName can be a proper foreign key constraint with cascade...

chriadam: I've now unified the codepaths so that every Key is a Secret in a Collection, so the constraints can be enforced appropriately.

chriadam:
I think this is now ready for review.

chriadam:
I've rewritten the history to chop the changes up into more manageable chunks as per IZh's suggestion.

Denis-Semakin:
LGTM

PR #7, opened 2017-12-01
[sailfish-secrets] Add SQLCipher-based EncryptedStoragePlugin. Contributes to JB#36797

chriadam:


chriadam:
Self approving.  Joona had a bunch of review comments (sent by email) which affect not just this PR but also the rest of the codebase, so I'll merge this one and then start to address all of those comments in a separate PR.

PR #8, opened 2017-12-05
Fix various review comments

chriadam:


jpetrell:
LGTM

chriadam:
Fixed another issue (nesting of library source trees) that Joona pointed out via email, and also fixed incorrect dbus introspection data which I noticed while adding the C API.  Merging.

PR #12, opened 2017-12-22
libsailfishsecretscrypto C API

chriadam:


chriadam:
So far, only the secrets api is finished (not crypto).  It uses (mostly) kernel style (and indentation, yuck, as enforced by Linus' lindent.sh, but with some minor modifications in cases where its output was clearly insane).

chriadam:
I haven't yet done the malloc checking etc.

chriadam:
Ok, Crypto API is now also included, and unit test has been added to exercise the common codepaths.  The coverage isn't great yet, of course.  And I haven't done full malloc result checking yet... I may leave that for a later time, and focus on higher-value work in the near future (e.g. documentation + examples for the C++ libs, Certificate support in the API, fleshing out the existing encryption plugins).

lib/SecretsCrypto/crypto.c
```diff
@@ -0,0 +1,1543 @@
...
+{
+	struct Sailfish_Crypto_Key_CustomParameter *param =
+	    (struct Sailfish_Crypto_Key_CustomParameter *)
+	    malloc(sizeof(struct Sailfish_Crypto_Key_CustomParameter));
+
+	if (parameter && parameterSize > 0) {
+		param->parameter = (unsigned char *)malloc(parameterSize);
+		memcpy(param->parameter, parameter, parameterSize);
+		param->parameterSize = parameterSize;
+		param->next = NULL;
```
Denis-Semakin: `param->next = NULL;` from here

lib/SecretsCrypto/crypto.c
```diff
@@ -0,0 +1,1543 @@
...
+	} else {
+		param->parameter = NULL;
+		param->parameterSize = 0;
+		param->next = NULL;
```
Denis-Semakin: `param->next = NULL;` from here

lib/SecretsCrypto/crypto.c
```diff
@@ -0,0 +1,1543 @@
...
+	}
+
```
Denis-Semakin: One may insert here

Denis-Semakin: But... Optimizer will do it for us I believe :)

lib/SecretsCrypto/crypto.c
```diff
@@ -0,0 +1,1543 @@
+/*
```
Denis-Semakin: To avoid reinvention of the wheel in future I would suggest to use Glib2's GList and etc to work around with linked lists, but now it's Ok

Denis-Semakin: Also, I strongly recommend to run this code under valgrind (or analogues) to check memory leaks at least ones.

lib/SecretsCrypto/secrets.c
```diff
@@ -0,0 +1,1000 @@
...
+
+void Sailfish_Secrets_Secret_FilterDatum_unref(
+		struct Sailfish_Secrets_Secret_FilterDatum *filter)
+{
+	if (filter) {
+		struct Sailfish_Secrets_Secret_FilterDatum *curr = filter;
+		struct Sailfish_Secrets_Secret_FilterDatum *next = filter->next
+		    ? filter->next : NULL;
+
+		while (curr) {
```
Denis-Semakin: Code duplication. I've seen this code in other files. Usually function or macro is created where all common code is implemented. Because what if you'll need to modify something, then you should change it in all these files.

Denis-Semakin: As I already mentioned I would suggest to use well-known frameworks to work with different kinds of linked lists.

lib/SecretsCrypto/secrets.c
```diff
@@ -0,0 +1,1000 @@
...
+	(void)source_object;
+	if (daemon_connection.p2pBus) {
+		free(daemon_connection.p2pAddr);
+		daemon_connection.p2pAddr = NULL;
+		g_dbus_proxy_new(
+		    daemon_connection.p2pBus,
+		    G_DBUS_PROXY_FLAGS_NONE,
+		    NULL,
+		    NULL, /* bus name */
+		    "/Sailfish/Secrets",
```
Denis-Semakin: Also. I would create a header file to store all string and numeric (returned values, any parameters, bus names, etc) constants in one place. It's convenient if one need to modify it.

lib/SecretsCrypto/secrets.c
```diff
@@ -0,0 +1,1000 @@
...
+Sailfish_Secrets_secretFromVariant(GVariant *variant)
+{
+	struct Sailfish_Secrets_Secret *secret = NULL;
+	GVariant *identVariant = NULL;
+	char *secretName = NULL;
+	char *collectionName = NULL;
+	GVariantIter *dataIter;
+	unsigned char datum = 0;
+	unsigned char *data =
+	    (unsigned char *)malloc(2048 * sizeof(unsigned char));
```
Denis-Semakin: What does 2048 means? What if we'll need to increase this value? Change it everywhere where we meet it?
`#define NEEDED_BYTES 2048` ???

lib/SecretsCrypto/secrets.c
```diff
@@ -0,0 +1,1000 @@
...
+	if (daemon_result) {
+		int resultCode = 0;
+		int errorCode = 0;
+		gchar *errorMessage = NULL;
+		g_variant_get(
+		    daemon_result, "((iis))",
+		    &resultCode,
+		    &errorCode,
+		    &errorMessage);
+		result = Sailfish_Secrets_Result_new(
```
Denis-Semakin: `result = Sailfish_Secrets_Result_new(resultCode, errorCode,
                                                              errorMessage);`
Better to align to `(` I suppose

lib/SecretsCrypto/secrets.c
```diff
@@ -0,0 +1,1000 @@
...
+{
+	return daemon_connection.secretsProxy != NULL;
+}
+
+int Sailfish_Secrets_connectToServer(
+		Sailfish_Secrets_connectToServer_callback callback,
+		void *callback_context)
+{
+	if (Sailfish_Secrets_connectedToServer()) {
+		return 2;	/* already connected */
```
Denis-Semakin: A magic constant... 
Why
```c
enum Sailfish_Secrets_Result_Code {
	Sailfish_Secrets_Result_Succeeded = 0,
	Sailfish_Secrets_Result_Pending = 1,
	Sailfish_Secrets_Result_Failed = 2
}
```
isn't used here?

Denis-Semakin: A magic constant

lib/Crypto/serialisation.cpp
```diff
@@ -513,19 +567,31 @@ const QDBusArgument &operator>>(const QDBusArgument &argument, Key::Identifier &
 
 QDBusArgument &operator<<(QDBusArgument &argument, const Key::FilterData &filterData)
 {
+    QVariantMap asv;
+    const QMap<QString,QString> fd = filterData;
```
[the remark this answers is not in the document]
chriadam: Thanks, I will test this and update.

lib/Crypto/serialisation.cpp
```diff
@@ -513,19 +567,31 @@ const QDBusArgument &operator>>(const QDBusArgument &argument, Key::Identifier &
...
+    for (QMap<QString,QString>::const_iterator it = fd.constBegin(); it != fd.constEnd(); it++) {
```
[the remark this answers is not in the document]
chriadam: Ok, I will update to use prefix

lib/Secrets/result.h
```diff
@@ -29,6 +29,8 @@ class Result {
         NoError = 0,
         UnknownError = 2,
         SerialisationError = 3,
+        DaemonError = 5,
```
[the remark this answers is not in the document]
chriadam: For symmetry with the Crypto::Result enum.  The first few values are fair generic, and I wanted to avoid potential confusion by aligning values where possible.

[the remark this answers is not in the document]
chriadam: The 1 value in particular is Pending from the Result::Code, which is why that one is "missing".

rpm/sailfish-secrets.spec
```diff
@@ -94,6 +94,38 @@ Requires:   libsailfishcrypto = %{version}-%{release}
 %description -n libsailfishcrypto-tests
 %{summary}.
 
+
+%package -n libsailfishsecretscrypto
```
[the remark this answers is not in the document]
chriadam: No runtime requirement, but does BuildRequire glib etc.  Note that if the daemon isn't installed/running, calls will simply fail.  I could add Requires:   %{name} = %{version}-%{release} to ensure the daemon is pulled in, I suppose that would be more useful.  I'll do that.

lib/SecretsCrypto/crypto.c
chriadam: Same for GHashTable etc too, I think, yes.  My inexperience with glib (and frustrations getting the DBus GVariant marshalling/demarshalling working) led me to doing it this suboptimal way.

I guess one question is: should the glib types by used in the header / API, or only in the .c impl?  If the latter, we can fix this later as an improvement.

chriadam: I ran the tst_secretscrypto unit test under valgrind.  The only leaks I saw were from glib's type registrations, and from glib internals in g_dbus_proxy_new.  That is, none of my manual allocations leaked, as far as I could tell (although I may have missed something, the amount of noise from glib was quite high..)

lib/SecretsCrypto/secrets.c
chriadam: Yes, I will try to deduplicate this code (and other things).  Not today (need to work on docs today), but when I revisit this PR, possibly early next week.

lib/SecretsCrypto/secrets.c
chriadam: Ok

lib/SecretsCrypto/secrets.c
chriadam: Urgh, I should update this to use GByteArray and the from_fixed_array accessor, as I did in crypto.c.  This entire block is suboptimal and error prone.  I will update.

lib/SecretsCrypto/secrets.c
chriadam: Will add to the private constants header when I add that one

lib/SecretsCrypto/secrets.c
chriadam: In many cases it's not possible to do so, while retaining the 80 column limit, due to kernel style's lengthy tab spacing.  Note that I ran lident over the source .c files, and made only minor adjustments where the output had very suboptimal readability.

chriadam:
Thanks for the review!  I will make the changes when I get a chance to revisit this PR (after working on documentation, so probably early next week).

lib/Secrets/result.h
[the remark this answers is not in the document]
vityafx: Is this symmetry actually useful? I mean, the enums are exactly for handling via names, not numbers. So, the key abstraction of name => number is useless then.

rpm/sailfish-secrets.spec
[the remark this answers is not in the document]
vityafx: What do you mean by it does not require anything at runtime? What about Qt packages, for example (QtCore, QDBus, and so on)? Also, as you mentioned, the daemon package.. The rpm packager will add those anyway, actually. I think so. We may check this after having an rpm file. But it is unclear from reading what it needs for runtime. At the very least case, even if you don't link to anything and don't use any dependency, you link to the libstdc++ (if you don't produce a static-linked executable). Just my opinion - I'd add anything for the reader to understand what it needs at a glance.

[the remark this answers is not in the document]
chriadam: This particular package is a glib-based C library.  It doesn't require anything from Qt at runtime.  (It does require qmake at build time, I suppose.)  It does link against glib / gio.  Those are included as BuildRequires.  Am I misunderstanding?

lib/Secrets/result.h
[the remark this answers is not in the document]
chriadam: I disagree, it is useful, especially considering the very low cost of ensuring the symmetry.
There are a variety of reasons, but the most compelling is that the (secrets) storage error code is reported in the crypto result, as some crypto operations (e.g. generateStoredKey) also interact with the secrets storage (indirectly).  So it's useful to have the "generic" error codes match each other.

rpm/sailfish-secrets.spec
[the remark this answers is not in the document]
vityafx: So if you dynamically linking to glib and gio, you should add both of them to the Require section, because rpm will check whether those are installed on the device.

chriadam:
Rebased, resolved conflicts.  Got the secrets API working again, but the crypto API is still WIP.

chriadam:
Superseded by the work by inz

PR #13, opened 2018-01-24
[sailfish-secrets] Add framework overview documentation. Contributes to JB#36797

chriadam:
This commit adds documentation for the Sailfish OS Secrets and Crypto
Framework, including overview documentation for the Sailfish OS Secrets
Library and the Sailfish OS Crypto Library.

It also changes the install path of C++ plugins and devel headers
for symmetry with the install path for QML plugins and documentation.

Contributes to JB#36797

### DIFF
--- a/doc/index.qdoc
+++ b/doc/index.qdoc
@@ -1,0 +1,173 @@
+/****************************************************************************************
+**
+** Copyright (C) 2018 Jolla Ltd.
+** Contact: Chris Adams <chris.adams@jollamobile.com>
+** All rights reserved.
+**
+****************************************************************************************/
+
+/*!
+\contentspage {Sailfish OS Secrets and Crypto Framework Contents}
+\page index.html
+
+\title Secret Data and Cryptographic Operations in Sailfish OS
+
+Applications running on highly-connected devices require the capability to
+perform advanced cryptographic operations, in order to protect the user's data,
+authenticate with remote services, verify the authenticity of other connected
+devices, and communicate securely.  The keys and certificates which are used
+to perform such operations also need to be stored securely on the device, in
+order to prevent malicious third-party applications from accessing those keys.
+
+The Sailfish OS Secrets and Crypto Framework provides applications with
+full-featured secret data storage and cryptography services.  Applications
+using the framework delegate cryptographic operations to a system service,
+and the framework has been designed to ensure that when used appropriately,
+secret key data need never enter the process address space of the application.
+Furthermore, if the cryptographic service provider plugin to the framework is
+backed by secure hardware, Trusted Execution Environment application, or
+Trusted Platform Module, then secret key data need never enter the normal
+"rich" execution environment.
+
+The framework also provides secure storage of secret data (including but not
+limited to keys, certificates, passwords, and other credentials) either in
+value-encrypted databases, block-encrypted databases, or hardware-backed
+secure storage.  Access to secret data requires user authentication which
+is provided either by the system device lock, a system-mediated authentication
+flow, an application-mediated authentication flow, or a custom authentication
+flow implemented by a framework extension plugin.
+
+\section1 Architecture of the Sailfish OS Secrets and Crypto Framework
+
+The framework consists of three major parts:
+
+\list
+\li The system daemon providing the services to clients
+\li Extension plugins allowing vendors or partners to extend the daemon's functionality
+\li Client libraries which allow applications to easily make use of the daemon's services
+\endlist
+
+In addition to those components, the system daemon makes use of other system
+services in order to properly implement the functionality and ensure that the
+security of the system is maintained, including the device lock service (in
+order to properly authenticate the user and unlock encrypted collections of
+secrets) and the access control service.
+
+\b{Note: the access control service is not yet integrated in the current
+version of the Sailfish OS Secrets and Crypto Framework, and as such the
+only access control mode which is supported is the Owner-Only mode!}
+
+If the application doesn't wish to make use of the system authentication flows,
+it may instead use the Sailfish OS Secrets Authentication UI Plugin to
+register an in-application authentication view with the system service.
+
+\b{Note: in-application authentication should be avoided unless your
+application is a trusted (vendor- or partner-supplied application)!}
+
+A diagram which shows the architecture follows:
+
+\badcode
+                                      +-------------+
+                                      |    Secure   |
+                                      |  Peripheral |
+                                      +-------------+
+                                          ^    ^
+                                          |    |
+                                    r-----'    '------,
+                                    |                 |
+                                 +--^---+-------+ +---^---+-------+
+                                 |crypto|plugins| |secrets|plugins|
+  +---------------------+        +------+-------+-+-------+-------+
+  |    Access Control   |<------<|                                |
+  |       Daemon        |>------>|                                |
+  +---------------------+  DBus  |                                |
+                                 |        sailfishsecretsd        |
+  +---------------------+        |                                |
+  |     Lock Screen     |<------<|                                |
+  |       Daemon        |>------>|                                |
+  | (SystemInteraction) |  DBus  |                                |
+  +---------------------+        +--------------------------------+
+                                     V ^            ^   ^
+                                     | |            |   |
+         r---------------------------' |            |   |
+         | .---------------------------'            |   |
+         | |               DBus                     |   |
+         V ^                                        |   |
+  +--------------------------+                      |   |
+  |     Sailfish Secrets     |                      |   |
+  |        UI Plugin         |                      |   |
+  | (ApplicationInteraction) |                      |   |
+  +--------------------------+  (Crypto API Call)   |   |
+  |                          |        DBus          |   |
+  |                          |>---------------------'   |
+  |         Client           |                          |
+  |        Application       |  (Secrets API Call)      |
+  |                          |        DBus              |
+  |                          |>-------------------------'
+  +--------------------------+
+\endcode
+
+\section1 Extending the Sailfish OS Secrets and Crypto Framework
+
+Device vendors and trusted partners may extend the Sailfish OS Secrets and
+Crypto Framework by installing extension plugins into the appropriate plugin
+location on the device.
+
+Please see the \l{Sailfish OS Crypto Library Overview} documentation and the
+\l{Sailfish OS Secrets Library Overview} documentation for more information
+on that topic.
+
+\section1 Using the Sailfish OS Secrets and Crypto Framework
+
+The functionality provided by the framework is exposed by three separate
+libraries: two Qt-based C++ libraries and one glib-based C library.
+Each of the libraries is simply a thin wrapper around IPC calls to the system
+daemon which provides the cryptography and secret storage services.  The
+C++ libraries provide significantly more syntactic sugar than the C library.
+A QML plugin also exists which provides applications with a simple-to-extend
+view which allows the application to implement an in-process authentication
+flow which inter-operates correctly with the system service daemon.
+
+\section2 The Sailfish OS Secrets Library
+
+This library provides client applications written in C++ (with Qt) with API to
+make use of the secrets storage services provided by the Sailfish OS Secrets
+and Crypto Framework.
+
+In-depth information can be found on the page about the
+\l{Sailfish OS Secrets Library Overview}.
+
+\section3 The Sailfish OS Secrets Authentication QML Plugin
+
+This plugin provides client applications written in C++ (with Qt and QML) with
+a QML component which allows in-process authentication flows to be implemented
+easily.  This type of authentication flow is useful when the the application
+uses Owner-Only access control semantics, and the user trusts the application
+with the authentication data.
+
+More information about how this component may be used in an application can be
+found at the page about the \l{Sailfish OS Secrets Authentication QML Plugin}.
+
+\b{Note: in general, system authentication flows should be used instead of
+in-process authentication flows, to ensure that the user need not trust the
+application with authentication data!}
+
+\section2 The Sailfish OS Crypto Library
+
+This library provides client applications written in C++ (with Qt) with API to
+make use of the cryptographic services provided by the Sailfish OS Secrets and
+Crypto Framework.
+
+In-depth information can be found on the page about the
+\l{Sailfish OS Crypto Library Overview}.
+
+\section2 The Sailfish OS Secrets/Crypto C Library
+
+This library provides client applications written in C (with glib) with API to
+make use of the secrets storage and cryptography services provided by the
+Sailfish OS Secrets and Crypto Framework.
+
+In-depth information can be found on the page about the
+\l{Sailfish OS Secrets/Crypto C Library Overview}.
+
+*/
--- a/lib/Crypto/doc/sailfish-crypto-overview.qdoc
+++ b/lib/Crypto/doc/sailfish-crypto-overview.qdoc
@@ -1,0 +1,218 @@
+/****************************************************************************************
+**
+** Copyright (C) 2018 Jolla Ltd.
+** Contact: Chris Adams <chris.adams@jollamobile.com>
+** All rights reserved.
+**
+****************************************************************************************/
+
+/*!
+\contentspage {Sailfish OS Crypto Library Contents}
+\page sailfish-crypto-overview.html
+
+\title Sailfish OS Crypto Library Overview
+
+The Sailfish OS Crypto Library provides applications with API to access the
+cryptography functionality offered by the system service daemon as part of the
+Sailfish OS Secrets and Crypto Framework.  Please see the Sailfish OS Secrets
+documentation for more in-depth documentation about the framework architecture,
+and broad overview of the features the framework provides.
+
+\section1 Using the Sailfish OS Crypto Library
+
+As described in the Sailfish OS Secrets and Crypto Framework overview
+documentation, client applications can use the Sailfish OS Crypto Library
+in order to make use of the cryptographic services provided by the
+Sailfish OS Secrets and Crypto Framework.
+
+This library provides client applications written in C++ (with Qt) with API to
+make use of the cryptographic services provided by the Sailfish OS Secrets and
+Crypto Framework.
+
+To make use of this library, applications should depend on the
+"sailfishcrypto.pc" pkgconfig file.
+
+e.g. in a qmake-based project:
+
+\code
+CONFIG += link_pkgconfig
+PKGCONFIG += sailfishcrypto
+\endcode
+
+\section2 Client API
+
+The client API consists of a variety of C++ classes which represent the
+inputs to cryptographic operations (including secret keys and certificates),
+the result of a cryptographic operation (that is, whether it succeeded or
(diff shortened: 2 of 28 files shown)


chriadam:
This provides overview documentation (and doc package installation etc), including usage snippets and information about plugins.  I still need to complete class and method API documentation.

doc/index.qdoc
```diff
@@ -0,0 +1,173 @@
...
+****************************************************************************************/
+
+/*!
+\contentspage {Sailfish OS Secrets and Crypto Framework Contents}
+\page index.html
+
+\title Secret Data and Cryptographic Operations in Sailfish OS
+
+Applications running on highly-connected devices require the capability to
+perform advanced cryptographic operations, in order to protect customer data,
```
jpetrell: User data maybe? Being customer is only one aspect of our users.

doc/index.qdoc
```diff
@@ -0,0 +1,173 @@
...
+\l{Sailfish OS Secrets Library Overview} documentation for more information
+on that topic.
+
+\section1 Using the Sailfish OS Secrets and Crypto Framework
+
+The functionality provided by the framework is exposed by three separate
+libraries: two Qt-based C++ libraries and one glib-based C library.
+Each of the libraries is simply a thin wrapper around IPC calls to the system
+daemon which provides the cryptography and secret storage services.  The
+C++ libraries provide significantly more syntax sugar than the C library.
```
jpetrell: syntax sugar -> syntactic sugar?

lib/Crypto/doc/sailfish-crypto-overview.qdoc
```diff
@@ -0,0 +1,218 @@
...
+// Set the identifier for the key.
+// This assumes the existence of an "ExampleCollection" secure storage
+// collection, in which the key will be stored.
+// See Sailfish::Secrets::SecretManager::createCollection().
+keyTemplate.setIdentifier(QStringLiteral("ExampleKey"),
+                          QStringLiteral("ExampleCollection"));
+
+// Ask the system service to generate and store the key securely.
+Sailfish::Crypto::CryptoManager cm;
+QDBusPendingReply<Sailfish::Crypto::Result, Sailfish::Crypto::Key>
```
jpetrell: If you look It is rare for DBus parts to leak outside the client-side Qt C++ wrapper API, usually the API provides all the necessary structures and signals to track the status of the requests. Might be a non-issue, just wondering.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
```diff
@@ -0,0 +1,218 @@
...
+// Set the key template metadata.
+Sailfish::Crypto::Key keyTemplate, symmetricKeyReference;
+keyTemplate.setAlgorithm(Sailfish::Crypto::Key::Aes256);
+keyTemplate.setOrigin(Sailfish::Crypto::Key::OriginDevice);
+keyTemplate.setBlockModes(Sailfish::Crypto::Key::BlockModeCBC);
+keyTemplate.setEncryptionPaddings(Sailfish::Crypto::Key::EncryptionPaddingNone);
+keyTemplate.setSignaturePaddings(Sailfish::Crypto::Key::SignaturePaddingNone);
+keyTemplate.setDigests(Sailfish::Crypto::Key::DigestSha256);
+keyTemplate.setOperations(Sailfish::Crypto::Key::Encrypt
+                         |Sailfish::Crypto::Key::Decrypt);
```
jpetrell: While repeating the namespaces is good for teaching the namespace structure it makes the example seem unnecessarily complex. I wish C++ supported Key key(algorithm=Aes256, origin=OriginDevice, blockmodes=BlockModeCBC, ..)  -style syntax, where the order of parameters didn't matter.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
```diff
@@ -0,0 +1,218 @@
...
+\code
+QByteArray ciphertext, plaintext = "Example plaintext data";
+QDBusPendingReply<Sailfish::Crypto::Result, QByteArray>
+        encryptReply = cm.encrypt(
+                plaintext,
+                symmetricKeyReference,
+                Sailfish::Crypto::Key::BlockModeCBC,
+                Sailfish::Crypto::Key::EncryptionPaddingNone,
+                Sailfish::Crypto::Key::DigestSha256,
+                Sailfish::Crypto::CryptoManager::DefaultCryptoStoragePluginName);
```
jpetrell: The style here is inconsistent with the above key template example, like discussed earlier Qt-style tries to avoid functions with a lot of parameters as it is not easy remember the order of passing (and in case of integers and strings there is a risk wrong parameters get passed without compiler complaining).

lib/Crypto/doc/sailfish-crypto-plugins.qdoc
```diff
@@ -0,0 +1,92 @@
...
+The first plugin provides implementations of common cryptography functionality
+by calling the OpenSSL library functions internally.
+
+The second plugin shares its cryptography implementation with the first plugin,
+but also provides block-level encrypted secure storage of collections of
+secrets, using the SQLCipher library for its database operations.  Thus it is
+both a \tt{Sailfish::Crypto::CryptoPlugin} and a
+\tt{Sailfish::Secrets::EncryptedStoragePlugin}.
+
+Note that neither of these plugins use TEE/TPM or other secure-hardware
```
jpetrell: Maybe TEE/TPM -> TEE / TPM. I wonder if we should write TEE/TPM in open form, not every knows what they mean. Though the subsequent phrase "or other secure-hardware" does give some context.

lib/Secrets/doc/sailfish-secrets-overview.qdoc
```diff
@@ -0,0 +1,242 @@
...
+secretReply.waitForFinished();
+if (secretReply.argumentAt<0>().code() == Sailfish::Secrets::Result::Failed) {
+    qWarning() << "Failed to retrieve secret:"
+               << secretReply.argumentAt<0>().errorMessage();
+} else {
+    qDebug() << "The secret data is:"
+             << secretReply.argumentAt<1>().data();
+}
+\endcode
+
```
jpetrell: At some point filtering example would be cool to have here also.

lib/Secrets/doc/sailfish-secrets-overview.qdoc
```diff
@@ -0,0 +1,242 @@
...
+\li \l{Sailfish::Secrets::Secret} represents a (possibly partial or reference) secret
+\li \l{Sailfish::Secrets::Secret::Identifier} consists of a secret name and optionally a collection name
+\li \l{Sailfish::Secrets::Result} represents the result (but not the output) of a secure storage operation
+\li \l{Sailfish::Secrets::SecretManager} provides an interface to the system secure storage service
+\endlist
+
+There also exist some API for dealing with authentication flows which require
+user interaction, where that user interaction is not mediated by the system
+(e.g., device lock). These APIs are listed here only for the sake of
+completeness as most applications should never use them:
```
jpetrell: I find this useful, but I guess it is noise for most app developers. Maybe it is good to be inclusive in overview documentation.

tests/plugins/testsqliteplugin/testsqliteplugin.pro
```diff
@@ -16,5 +16,5 @@ HEADERS += \
 SOURCES += \
     $$PWD/../../../plugins/sqliteplugin/plugin.cpp
 
-target.path=/usr/lib/sailfish/secrets/
+target.path=/usr/lib/Sailfish/Secrets/
```
jpetrell: Not a biggie, but changing installation paths in documentation commit is a bit funny.

lib/Secrets/doc/sailfish-secrets.cpp
```diff
@@ -0,0 +1,17 @@
+namespace Sailfish {
+namespace Secrets {}
+}
+
+/*!
+  \namespace Sailfish
+  \brief This namespace provides various APIs to interact with the Sailfish OS system.
+ */
```
jpetrell: Is there any way to avoid duplicating this code and comments between secrets and crypto?

lib/Secrets/doc/sailfish-secrets-plugins.qdoc
```diff
@@ -0,0 +1,118 @@
...
+The following snippet shows an example of creating a collection of secrets
+in an encrypted database managed by the default encrypted storage plugin:
+
+\code
+Sailfish::Secrets::SecretManager sm;
+QDBusPendingReply<Sailfish::Secrets::Result>
+        collectionReply = sm.createCollection(
+                QLatin1String("ExampleCollection"),
+                Sailfish::Secrets::SecretManager::DefaultEncryptedStoragePluginName,
+                Sailfish::Secrets::SecretManager::DefaultEncryptedStoragePluginName,
```
jpetrell: Two identical parameter lines?

jpetrell:
Overall LGTM.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
vityafx: @jpetrell It can be solved by putting `using namespace Sailfish::Crypto;` into the scope. Then it will be just `Key::Aes256`, for example. The other way is to make a typedef/using: `using Key = Sailfish::Crypto::Key;` and then just `Key::Aes256`.

Also, the use of knowing the exact fully qualified path is under doubts. Boost has namespaces too, but everyone and even they themselves use aliases to their types, like `using po = boost::program_options;`. People will know the path but there are doubts anyone would use the full path in each line (saves key presses and time at the very least.

Also, by having these unspoken rules where namespace names are in UpperCamelCase it is impossible to know what is the type and what is the namespace for the developer: that is why everyone uses lower camel case or lower case with underscores for the namespaces. Qt uses UpperCamelCase only for the types, not namespaces. I have tried to find a rule where you could possibly see it but there is nothing like that. And I don't remember that it was something like that there.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
vityafx: I'd never leak outside the transport. I'd even hide everything so that the user would never know that the data is even passed somewhere. The user must not know that. My opinion. Also, what would the user have if we suddenly decide to change the DBus to anything else? Or just to another dbus library? All the user code will need to be changed. Do we really want it? Also, as I told before, it makes the user code more complicated since the user must everytime to unpack the result. Useless work. But it will be everywhere in the user code. Probably, it is so useless work that everyone will write his own abstractions (wrappers) around that work if it is possible. I'd do that. I'd do that even due to another reasons such as abstractions of future changes of the API (eg we are moving from dbus to anything else which need not to be unpacked). So, I'd simply not leak outside anything which is not related to the purpose of the framework.

No offense please, I hope I was polite enough. Want to help. <3

lib/Crypto/doc/sailfish-crypto-overview.qdoc
vityafx: iOS provides a map where you must put all this [information](https://www.andyibanez.com/using-ios-keychain/):

```objective-c
    //Let's create an empty mutable dictionary:
    NSMutableDictionary *keychainItem = [NSMutableDictionary dictionary];
     
    NSString *username = self.nwUsernameTxtFld.text;
    NSString *password = self.passwordTxtFld.text;
    NSString *website = @"<a class="vglnk" href="http://www.myawesomeservice.com" rel="nofollow"><span>http</span><span>://</span><span>www</span><span>.</span><span>myawesomeservice</span><span>.</span><span>com</span></a>";
     
    //Populate it with the data and the attributes we want to use.
     
    keychainItem[(__bridge id)kSecClass] = (__bridge id)kSecClassInternetPassword; // We specify what kind of keychain item this is.
    keychainItem[(__bridge id)kSecAttrAccessible] = (__bridge id)kSecAttrAccessibleWhenUnlocked; // This item can only be accessed when the user unlocks the device.
    keychainItem[(__bridge id)kSecAttrServer] = website;
    keychainItem[(__bridge id)kSecAttrAccount] = username;
```

Here we will never have a mistake. We could also have a struct for these purposes.

The actual processing of this data is performed by only one function call with only one argument - this dictionary. It returns boolean result. Simple as possible. We may simply look what other systems provide and do something similar.

vityafx:
How do we handle a situation?

1. User installs the app.
2. App stores some data in the Secrets API.
3. App is being deleted.

What will be with the secrets it stored? For example, iOS can handle this, if we delete app and reinstall it after, it will ask for a password again.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
```diff
@@ -0,0 +1,218 @@
...
+\section1 Extending the Sailfish OS Secrets and Crypto Framework with Crypto Plugins
+
+The Sailfish OS Crypto Library also provides a plugin base-class which may be
+extended by device vendors or trusted partners to allow them to build plugins
+to extend the Sailfish OS Secrets and Crypto Framework with additional
+cryptography functionality (for example, supporting different algorithms or
+block modes, or performing the operations via a Trusted Execution Environment
+application rather than in-process in the rich application process).
+
+The \l{Sailfish::Crypto::CryptoPlugin} class should be extended in order to
```
vityafx: Did we decide how we handle custom algorithms, which are not listed in the enumeration?

lib/Crypto/doc/sailfish-crypto-overview.qdoc
chriadam: No offense taken, and you're right.  However, given that documentation became a high priority item, I thought it best to write docs for what currently exists immediately, rather than spending potentially large amount of time designing and implementing a different client API - because (and here's the important thing) I believe that the MOST IMPORTANT thing we have to resolve right now is: does the API support the required features, and are the parameters to the operations correct.  Documenting the current API helps us to resolve these questions (I hope...).

Next week I will take your suggestion and start prototyping a client API (which internally wraps around the current Secret/CryptoManager APIs which will become private API) which hides the transport and also uses some form of per-parameter accessor/mutator request data object.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
chriadam: Qt does use camel-cased namespaces, e.g. "QPatternist", "QtAndroid", "QtMobility", "Simulator", "QDBus", "QMdi", etc etc.

Yes, the developer can have "using namespace" however that can cause issues when using both the Crypto and the Secrets API, since some type names are shared (which is a major reason to use namespaces, avoiding polluting the global namespace).  Yes, using full symbol naming makes the examples more verbose BUT it also makes them more precise.  We should assume that our client developers are intelligent and able to figure out how to improve the readability of their own code in the manner they see fit.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
chriadam: Yes, I will do something similar for the client API when I create the wrapper, with per-request-type request objects.

lib/Crypto/doc/sailfish-crypto-plugins.qdoc
chriadam: I think it's fine, this is a secure storage and cryptography API...

lib/Secrets/doc/sailfish-secrets-overview.qdoc
chriadam: I'd prefer not to leave out details, since some people might look at the header list and think "but what about these exported classes?  what are they for?  do I need them?" -- better to clearly state that they are not useful in the majority of circumstances, and let the developer move on, IMO.

lib/Secrets/doc/sailfish-secrets-plugins.qdoc
chriadam: Yes, encrypted storage plugin provides both encryption and storage, thus it is the same parameter for both.  I could add a comment here noting that.

lib/Secrets/doc/sailfish-secrets.cpp
chriadam: Possibly, I could split out "sailfish.cpp" and move it to the top level doc/ directory, which contains just the namespace definition, and then include that one from "sailfish-secrets.cpp" and "sailfish-crypto.cpp" in the per-module doc/ directories.

doc/index.qdoc
```diff
@@ -0,0 +1,173 @@
...
+A QML plugin also exists which provides applications with a simple-to-extend
+view which allows the application to implement an in-process authentication
+flow which inter-operates correctly with the system service daemon.
```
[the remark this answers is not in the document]
chriadam: Only in the application.  The system authentication flow UI is defined by the system, not the application.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
chriadam: Not yet, it is an open question.  In my opinion, we could add another enum PluginCustomAlgorithm (perhaps with value 1, i.e. next after UnknownAlgorithm), and document it to say that if that algorithm is specified, a plugin-specific CustomParameter must be provided to tell the plugin which algorithm to use.  Does anyone object to that idea, or have any other suggestions?

chriadam:
AFAIK we don't have "real" application management (i.e., some system service which handles installation/uninstallation of applications, and sends appropriate signals to other services to notify of such events).  This means that currently, e.g. if an application creates a specific calendar, or account, and then the application is deleted, then those pieces of data will not be deleted by Sailfish OS (the user will have to delete them manually).

I think we do need some form of well-defined resource policy (e.g. delete application-owned resources on application uninstallation) but before we can implement that properly, we need a proper access control framework (i.e. to be able to determine which resources are owned by / used by ONLY that application, etc).

lib/Crypto/doc/sailfish-crypto-overview.qdoc
vityafx: Good! Okay.:)

lib/Crypto/doc/sailfish-crypto-overview.qdoc
vityafx: So I’d use short type alias.

lib/Crypto/doc/sailfish-crypto-overview.qdoc
vityafx: Nice!

vityafx:
@chriadam I think before doing that we also must implement the application id functionality.

doc/index.qdoc
chriadam: done

doc/index.qdoc
chriadam: done

lib/Secrets/doc/sailfish-secrets.cpp
chriadam: done

chriadam:
Merging this one for now, we can revisit documentation after I've made the client API changes (wrapper API, fixes as per feedback, and added functionality like CSPRNG, key data generation, requesting passphrase data from user, etc).

PR #14, opened 2018-01-30
Improve the client API interface

chriadam:


chriadam:
WIP!  Not ready for review.

This PR adds per-request wrapper objects, providing a nicer client API without leaking transport (DBus) information into the interface.

Still lots to do, e.g. move the CryptoManager methods into a private class and make the Request types friend classes of the CryptoManager so they can access that private class; also need to add unit tests for the new request objects, etc.  Then I need to do the same for the Secrets API.  And add/update documentation.

chriadam:
This is now ready for review.

lib/Crypto/cryptodaemonconnection_p.h
```diff
@@ -8,35 +8,52 @@
...
     bool connect();
+    QDBusConnection *connection();
+    QDBusInterface *createInterface(const QString &objectPath,
+                                    const QString &interface,
+                                    QObject *parent = Q_NULLPTR);
+
+    static void registerDBusTypes();
 
-public Q_SLOTS:
+signals:
```
vityafx: Some incosistency: you use `Q_SIGNALS` most of time but here is `signals:`.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -302,15 +293,19 @@ void tst_cryptorequests::storedKeyRequests()
...
+    Sailfish::Secrets::CreateCollectionRequest ccr(&sm);
+    ccr.setCollectionLockType(Sailfish::Secrets::CreateCollectionRequest::DeviceLock);
+    ccr.setCollectionName(QLatin1String("tstcryptosecretsgcsked"));
+    ccr.setStoragePluginName(DEFAULT_TEST_CRYPTO_STORAGE_PLUGIN_NAME);
+    ccr.setEncryptionPluginName(DEFAULT_TEST_CRYPTO_STORAGE_PLUGIN_NAME);
+    ccr.setAuthenticationPluginName(IN_APP_TEST_AUTHENTICATION_PLUGIN);
+    ccr.setDeviceLockUnlockSemantic(Sailfish::Secrets::SecretManager::DeviceLockKeepUnlocked);
+    ccr.setAccessControlMode(Sailfish::Secrets::SecretManager::OwnerOnlyMode);
+    ccr.setUserInteractionMode(Sailfish::Secrets::SecretManager::ApplicationInteraction);
+    ccr.startRequest();
```
vityafx: As a user, I may simply want everything work in blocking mode. Since we don't have such an option, I'd also provide a `performRequest()` method to avoid writing "waiting" macros all the time in the user code.

lib/Secrets/doc/sailfish-secrets-overview.qdoc
```diff
@@ -203,15 +199,14 @@ a collection is very similar to that for deleting a secret, so only the
...
-                Sailfish::Secrets::Secret::Identifier("ExampleSecret",
-                                                      "ExampleCollection"),
-                SecretManager::ApplicationInteraction);
-deleteReply.waitForFinished();
-if (deleteReply.argumentAt<0>().code() == Sailfish::Secrets::Result::Failed) {
+Sailfish::Secrets::DeleteCollectionRequest dcr(&sm);
+dcr.setCollectionName(QLatin1String("ExampleCollection"));
+dcr.setUserInteractionMode(Sailfish::Secrets::SecretManager::SystemInteraction);
+dcr.startRequest();
+dcr.waitForFinished();
```
vityafx: I have already told about `performRequest`, I see there is a method for waiting, but it is still better from user-point of view to have a separate method, for example, `performRequest()` which simply calls `startRequest()` and then `waitForFinished()`.

vityafx:
I saw a lot of similar code so I simply looked over 5 of them or so and began looking at next commits. Am I correct saying this PR has mostly the transport abstraction (moving to request pattern)? Or did I miss something?

lib/Crypto/cryptodaemonconnection_p.h
chriadam: Thanks, fixed

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
chriadam: This is true.  Let's do that in a separate PR, as we can add that later in a BC/SC manner, and the other API issues are probably more important to fix first.

chriadam:
Yes, that is all this one does.  If you're happy with it, I will merge (as the functional API changes will require this patch set to be merged, first)

PR #16, opened 2018-02-05
[sailfish-crypto] Allow the client to limit retrieve key data. Contributes to JB#40058

chriadam:
Previously, the StoredKeyRequest would retrieve all of the data
associated with a key entry, including metadata, public key data, and
private key data.  This commit allows the client to specify which
of these portions of data they wish to retrieve.

This allows, for example, the client to retrieve just the public
key data portion, allowing the client to limit potential secret
key data leakage in the case where they suspect that the secrecy
of their process's memory address space may be compromised.

Contributes to JB#40058

### DIFF
--- a/daemon/CryptoImpl/cryptorequestprocessor.cpp
+++ b/daemon/CryptoImpl/cryptorequestprocessor.cpp
@@ -29,6 +29,33 @@ namespace {
         }
         return retn;
     }
+
+    void nullifyKeyFields(Sailfish::Crypto::Key *key, Sailfish::Crypto::StoredKeyRequest::KeyComponents keep) {
+        // This method is called for keys stored in generic secrets storage plugins.
+        // Null-out fields if the client hasn't specified that they be kept.
+        // Note that by default we treat CustomParameters as PublicKeyData.
+        if (!(keep & Sailfish::Crypto::StoredKeyRequest::MetaData)) {
+            key->setIdentifier(Sailfish::Crypto::Key::Identifier());
+            key->setOrigin(Sailfish::Crypto::Key::OriginUnknown);
+            key->setAlgorithm(Sailfish::Crypto::Key::AlgorithmUnknown);
+            key->setBlockModes(Sailfish::Crypto::Key::BlockModeUnknown);
+            key->setEncryptionPaddings(Sailfish::Crypto::Key::EncryptionPaddingUnknown);
+            key->setSignaturePaddings(Sailfish::Crypto::Key::SignaturePaddingUnknown);
+            key->setDigests(Sailfish::Crypto::Key::DigestUnknown);
+            key->setOperations(Sailfish::Crypto::Key::OperationUnknown);
+            key->setFilterData(Sailfish::Crypto::Key::FilterData());
+        }
+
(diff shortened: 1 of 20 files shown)


lib/Crypto/storedkeyrequest.h
```diff
@@ -26,15 +26,28 @@ class SAILFISH_CRYPTO_API StoredKeyRequest : public Sailfish::Crypto::Request
...
     Q_PROPERTY(Sailfish::Crypto::Key::Identifier identifier READ identifier NOTIFY identifierChanged)
+    Q_PROPERTY(KeyComponents keyComponents READ keyComponents WRITE setKeyComponents NOTIFY keyComponentsChanged)
     Q_PROPERTY(Sailfish::Crypto::Key storedKey READ storedKey NOTIFY storedKeyChanged)
 
 public:
+    enum KeyComponent {
+        NoData          = 0,
+        MetaData        = 1,
+        PublicKeyData   = 2,
+        SecretKeyData   = 4
```
vityafx: The scenario when we combine `PublicKeyData` and `SecretKeyData` is possible, am I right?

daemon/CryptoImpl/cryptorequestprocessor.cpp
```diff
@@ -29,6 +29,33 @@ namespace {
         }
         return retn;
     }
+
+    void nullifyKeyFields(Sailfish::Crypto::Key *key, Sailfish::Crypto::StoredKeyRequest::KeyComponents keep) {
```
vityafx: I am not sure I understand why the `nullifyKeyFields` exists more than once in the code. My thought that it would work like this:

1. We create a crypto request and send it.
2. The daemon receives it and asks the appropriate plugin to get the key.
3. The plugin returns the key data.
4. The code in the crypto request processor calls `nullifyKeyFields`.

But I see this function three times in this PR. Did I miss the conception?

lib/Crypto/storedkeyrequest.h
chriadam: Of course, they're flags.  This is explicitly unit tested also.

daemon/CryptoImpl/cryptorequestprocessor.cpp
chriadam: The reason is that, in the case where the crypto plugin is also a storage plugin, we never want the secret key data to leak back to the daemon address space, in the case where the client only requests PublicKeyData.  Thus, the plugin (and not the daemon) should implement the field nullification.  While in our case the plugin is loaded in-process it probably doesn't make much difference, but it certainly does for TEE/TPM case.

The "fallback" case (where the daemon performs the nullification) is only applied where the key is stored as a "normal" secret in a secrets storage plugin (rather than crypto storage plugin).

However, the third case (opensslcryptoplugin) should NOT have that method, since it is not a storage plugin.  I copy pasted it there incorrectly, and I will remove that copy of it.

vityafx: @chriadam Good, thanks for the explanation.

lib/Crypto/storedkeyrequest.h
vityafx: Okay, thanks.

chriadam:
Thanks, merging now.

PR #17, opened 2018-02-06
[sailfish-crypto] Add CSPRNG functionality. Contributes to JB#40058

chriadam:


lib/Crypto/generaterandomdatarequest.h
```diff
@@ -0,0 +1,76 @@
...
+#include <QtCore/QString>
+#include <QtCore/QByteArray>
+
+namespace Sailfish {
+
+namespace Crypto {
+
+class CryptoManager;
+
+class GenerateRandomDataRequestPrivate;
```
vityafx: Why not to put it inside the `GenerateRandomDataRequest` class itself?

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
```diff
@@ -18,20 +20,83 @@
...
+        QByteArray *randomData)
+{
+    Q_UNUSED(callerIdent)
+
+    if (csprngEngineName != GenerateRandomDataRequest::DefaultCsprngEngineName) {
+        return Result(Result::CryptoPluginRandomDataError,
+                      QLatin1String("The OpenSSL crypto plugin doesn't currently support other RNG engines")); // TODO!
+    }
+
+    if (numberBytes >= 4097 || numberBytes <= 0) {
```
vityafx: An unsigned integer can't be lower than `0`. I'd rewrite it as `if (!numberBytes || numberBytes > MAX_NUMBER_OF_BYTES)`.

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
```diff
@@ -18,20 +20,83 @@
...
+        return Result(Result::CryptoPluginRandomDataError,
+                      QLatin1String("The OpenSSL crypto plugin can only generate up to 4096 bytes of random data at a time"));
+    }
+
+    int nbytes = numberBytes;
+    unsigned char buf[4096];
```
vityafx: I'd put this `4096` and `4097` above into a single named constant.

plugins/sqlcipherplugin/cryptoplugin.cpp
```diff
@@ -20,11 +21,69 @@
...
+        QByteArray *randomData)
+{
+    Q_UNUSED(callerIdent)
+
+    if (csprngEngineName != Sailfish::Crypto::GenerateRandomDataRequest::DefaultCsprngEngineName) {
+        return Sailfish::Crypto::Result(Sailfish::Crypto::Result::CryptoPluginRandomDataError,
+                                        QLatin1String("The SQLCipher crypto plugin doesn't currently support other RNG engines")); // TODO!
+    }
+
+    if (numberBytes >= 4097 || numberBytes <= 0) {
```
vityafx: The same as above.

plugins/sqlcipherplugin/cryptoplugin.cpp
```diff
@@ -20,11 +21,69 @@
...
+        return Sailfish::Crypto::Result(Sailfish::Crypto::Result::CryptoPluginRandomDataError,
+                                        QLatin1String("The SQLCipher crypto plugin can only generate up to 4096 bytes of random data at a time"));
+    }
+
+    int nbytes = numberBytes;
+    unsigned char buf[4096];
```
vityafx: The same as above.

plugins/sqlcipherplugin/cryptoplugin.cpp
```diff
@@ -20,11 +21,69 @@
...
+    if (RAND_bytes(buf, nbytes) != 1) {
+        return Sailfish::Crypto::Result(Sailfish::Crypto::Result::CryptoPluginRandomDataError,
+                                        QLatin1String("The SQLCipher crypto plugin failed to generate the random data"));
+    }
+
+    *randomData = QByteArray((char*)buf, nbytes);
```
vityafx: I'd prefer getting rid of C-style casts by providing the appropriate g++ compiler warning flag.

lib/Crypto/generaterandomdatarequest.h
chriadam: Private implementation class allows us to make changes later which would otherwise be binary incompatible.  I will do similar pimpl-isation of the rest of the API in the near future.

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
chriadam: Indeed, will do

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
chriadam: Yep, will use the MAX_NUMBER_OF_BYTES define as you suggested

plugins/sqlcipherplugin/cryptoplugin.cpp
chriadam: I will use a C++ reinterpret cast instead

lib/Crypto/cryptomanager.cpp
```diff
@@ -76,6 +76,50 @@ CryptoManagerPrivate::getPluginInfo()
...
+{
+    if (!m_interface) {
+        return QDBusPendingReply<Result, QByteArray>(
+                    QDBusMessage::createError(QDBusError::Other,
+                                              QStringLiteral("Not connected to daemon")));
+    }
+
+    QDBusPendingReply<Result> reply
+            = m_interface->asyncCallWithArgumentList(
+                "seedRandomDataGenerator",
```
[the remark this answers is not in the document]
chriadam: Will do

daemon/CryptoImpl/crypto.cpp
```diff
@@ -323,6 +365,53 @@ void Daemon::ApiImpl::CryptoRequestQueue::handlePendingRequest(
             }
             break;
         }
+        case GenerateRandomDataRequest: {
+            qCDebug(lcSailfishCryptoDaemon) << "Handling GenerateRandomDataRequest from client:" << request->remotePid << ", request number:" << request->requestId;
+            QByteArray randomData;
+            quint64 numberBytes = request->inParams.size() ? request->inParams.takeFirst().value<quint64>() : 0;
+            QString csprngEngineName = request->inParams.size() ? request->inParams.takeFirst().value<QString>() : QString();
+            QString cryptosystemProviderName = request->inParams.size() ? request->inParams.takeFirst().value<QString>() : QString();
+            Result result = m_requestProcessor->generateRandomData(
```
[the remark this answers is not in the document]
chriadam: We need to fix the way we marshal parameters internally a bit better, definitely (I think we need per-request-type structs, to avoid this list-style procedure).  But its important to note that the fallback cases here should never occur in practice (so we should add Q_ASSERTs to check them), the default constructed values are there solely to avoid a crash in case we missed some codepath in our unit tests which gets hit in the wild.  Yes, we could and should return an error immediately; but let's do that at the same time we fix the marshalling, IMO.

chriadam:
Updated and rebased, let me know if you're happy for me to merge this one

daemon/CryptoImpl/crypto.cpp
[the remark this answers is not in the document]
vityafx: Ok.

lib/Crypto/generaterandomdatarequest.h
vityafx: I know what is PIMPL. I ask why we could not put the line `class GenerateRandomDataRequestPrivate` inside the `class GenerateRandomDataRequest {` so it would look like:

```c++
class GenerateRandomDataRequest
{
//....
private:
    class GenerateRandomDataRequestPrivate;
};
```

vityafx:
@chriadam Could you comment "Probably it would also be good to add generateRandomNumber method which accept limits." please?

chriadam:
Added TODO comment, and rebased.  Will merge now.  Thanks!

PR #18, opened 2018-02-06
[sailfish-crypto] Improve the encrypt/decrypt client API. Contributes to JB#40058

chriadam:
Add support for user-supplied initialisation vector data, and remove
the unnecessary digest parameter.

Contributes to JB#40058

### DIFF
--- a/daemon/CryptoImpl/crypto.cpp
+++ b/daemon/CryptoImpl/crypto.cpp
@@ -239,10 +239,10 @@ void Daemon::ApiImpl::CryptoDBusObject::verify(
 
 void Daemon::ApiImpl::CryptoDBusObject::encrypt(
         const QByteArray &data,
+        const QByteArray &iv,
         const Key &key,
         Key::BlockMode blockMode,
         Key::EncryptionPadding padding,
-        Key::Digest digest,
         const QString &cryptosystemProviderName,
         const QDBusMessage &message,
         Result &result,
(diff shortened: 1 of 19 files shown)


daemon/CryptoImpl/crypto.cpp
```diff
@@ -508,19 +508,19 @@ void Daemon::ApiImpl::CryptoRequestQueue::handlePendingRequest(
             qCDebug(lcSailfishCryptoDaemon) << "Handling EncryptRequest from client:" << request->remotePid << ", request number:" << request->requestId;
             QByteArray encrypted;
             QByteArray data = request->inParams.size() ? request->inParams.takeFirst().value<QByteArray>() : QByteArray();
+            QByteArray iv = request->inParams.size() ? request->inParams.takeFirst().value<QByteArray>() : QByteArray();
             Key key = request->inParams.size() ? request->inParams.takeFirst().value<Key>() : Key();
             Key::BlockMode blockMode = request->inParams.size() ? request->inParams.takeFirst().value<Key::BlockMode>() : Key::BlockModeUnknown;
             Key::EncryptionPadding padding = request->inParams.size() ? request->inParams.takeFirst().value<Key::EncryptionPadding>() : Key::EncryptionPaddingUnknown;
-            Key::Digest digest = request->inParams.size() ? request->inParams.takeFirst().value<Key::Digest>() : Key::DigestUnknown;
             QString cryptosystemProviderName = request->inParams.size() ? request->inParams.takeFirst().value<QString>() : QString();
```
vityafx: Can the `m_requestProcessor` actually do anything useful with all of these parameters evaluated to defaults?

chriadam: As mentioned in the other PR, the default parameter fallbacks are purely to prevent a crash, and should never occur in practice (if the parameter doesn't exist, it would indicate that we've marshalled the asynchronous request wrongly).  The "real" fix for this is to fix the marshalling to use per-request-type structs instead of QVariantList -- but that should be done later once we've finished the API improvements, IMO.

vityafx: Ok.

chriadam:
Rebased, merging.

PR #19, opened 2018-02-09
CipherRequest

chriadam:


daemon/CryptoImpl/crypto_p.h
```diff
@@ -149,6 +149,50 @@ class CryptoDBusObject : public QObject, protected QDBusContext
...
+    "          <annotation name=\"org.qtproject.QtDBus.QtTypeName.In2\" value=\"Sailfish::Crypto::Key::BlockMode\" />\n"
+    "          <annotation name=\"org.qtproject.QtDBus.QtTypeName.In3\" value=\"Sailfish::Crypto::Key::EncryptionPadding\" />\n"
+    "          <annotation name=\"org.qtproject.QtDBus.QtTypeName.In4\" value=\"Sailfish::Crypto::Key::SignaturePadding\" />\n"
+    "          <annotation name=\"org.qtproject.QtDBus.QtTypeName.In5\" value=\"Sailfish::Crypto::Key::Digest\" />\n"
+    "          <annotation name=\"org.qtproject.QtDBus.QtTypeName.Out0\" value=\"Sailfish::Crypto::Result\" />\n"
+    "      </method>\n"
+    "      <method name=\"updateCipherSessionAuthentication\">\n"
+    "          <arg name=\"authenticationData\" type=\"ay\" direction=\"in\" />\n"
+    "          <arg name=\"cryptosystemProviderName\" type=\"s\" direction=\"in\" />\n"
+    "          <arg name=\"cipherSessionToken\" type=\"u\" direction=\"in\" />\n"
```
monich: If this is a wrapper around a block cipher then this looks slightly weird - why to provide authenticationData, cryptosystemProviderName and cipherSessionToken for every block? Wouldn't a file descriptor be enough?

monich: Not to mention that the whole thing looks like a huge overhead (what's the advantage of doing block cipher this way?).

chriadam: The cipher session authentication data isn't provided for every block, it's the initial data update to the cipher when in GCM mode, which allows generation of a specific tag allowing verification of the decrypted stream.

When you say "wouldn't a file descriptor be enough" do you mean a socket descriptor?  Perhaps that would indeed be better - patches welcome.  It won't affect the client API so any such implementation change will be BC/SC compatible from client perspective.

monich: Either a [socketpair](http://man7.org/linux/man-pages/man2/socketpair.2.html) or a pair of pipes would do if we are talking about a symmetric block cipher (when the size of the encrypted or decrypted data exactly equals the size of the input). D-Bus can pass file descriptors between the processes. Writing/reading a file descriptor would be significantly more efficient (although still significantly slower than doing everything without any IPC involved).

Feel free to ignore my comments - it could be that I simply no longer understand what this is about.

chriadam: No, you are correct that we're doing a bunch of data passing via the p2p dbus socket here, which means that there's a bunch of marshaling and demarshaling etc - so no doubt there is quite a lot of overhead.

And if it's simple to fix, we should do that.  With the way the API is designed, we can swap out the implementation details as needed.  I don't think those implementation changes should block this change from being merged, though, as such changes can be done in BC/SC way.

FWIW I just implemented "batch mode" for Cipher requests, which greatly improves performance.  I get approximately 13 MB per second encryption and decryption throughput on my L500D device, which probably isn't too far off maximal throughput (judging from https://wiki.openwrt.org/doc/howto/benchmark.openssl which shows motorola droid xt910 achieving 16 MB/sec).

Issue #5, opened 2017-11-28
Use sailfishsecretsd as a GPG agent, is it a good idea?


dcaliste:
From what I've understand of the README, this daemon will provide a secured and unified way to access crypto operations. That's great. There is already a (restricted) framework, but widely used, doing that namely GPG. So I'm brainstorming on how to integrate the two projects.

For reference here is how GPG operations are handled:

* there is a daemon-like process called GPG agent running (or started on demand). This daemon is playing an intermediate role between the application and possible user interaction.
* when a program is doing a crypto operation (like sign for instance), using either gpg command-line program or gpgme API, a IPC connection is opened with the GPG agent if some user interaction is required.
* then, the GPG agent may start a pinentry program and ask it via another IPC the passphrase. Then the passphrase is returned to the application via IPC.
* finally gpgme internal handle the IPC with the GPG program actually doing the operation and returned data to the application through IPC.

Using the default GPG agent is not convenient in Sailfish for these reasons:
* it is starting a new process (the pinentry) each time a user interaction is required.
* it seems to me that it's difficult to integrate this new process into lipstick: how to tell lipstick that this process would have no window, should be modal and displayed as a system dialog?

So my question, why not make the brand new sailfishsecretsd implements a GPG agent? I've done this already as a test for GPG signature in mail, it's not complicated. How would this be done? Which plugin API should be used?

chriadam:
First, yes you're correct about what sailfish-secrets provides.  I agree that it is a good idea to ensure that sailfish-secrets fulfills the needs of all clients, including doing what GPG does currently.

In regards to the user-interaction flows, this is a core consideration for our platform.  Specifically, we are currently adding support to devicelock daemon to allow clients to request devicelock to perform a "generic" user interaction flow (where the actual UI will be shown by lipstick).  In addition to those "system-mediated" UI flows, the secrets API also supports "in-app" flows (in fact, those are already implemented and can be tested right now).

[...]

PR #3, opened 2017-11-16
A patch-set to start work on CryptoToki plugin

Denis-Semakin:
This plugin is aimed to provide a high level interface to interact with Cryto Token USB devices supported PKSC#11 standard.

PR #1, opened 2017-11-09
Add QtTest to build requires for test packages

aleksey-mikhaylov:
Signed-off-by: Aleksey Mikhaylov <a.mikhaylov@omprussia.ru>

PR #21, opened 2018-02-21
Add an error message when plugin cannot be loaded.

dcaliste:
It helps to find the root cause of a plugin not loading, like missing symbols…

### DIFF
--- a/daemon/CryptoImpl/cryptorequestprocessor.cpp
+++ b/daemon/CryptoImpl/cryptorequestprocessor.cpp
@@ -13,6 +13,7 @@
 
 #include "util_p.h"
 #include "logging_p.h"
+#include "plugin_p.h"
 
 #include <QtCore/QDir>
 #include <QtCore/QPluginLoader>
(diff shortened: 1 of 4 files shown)


vityafx:
Could you provide an example of how useful it is? I mean, any output of missing symbols? I am very interested in this PR and what is actually does.

dcaliste:
You are developing a plugin and you forgot to link it with dependencies in the pro file. It will compiles but there will be unresolved symbols on load. Or you forgot to implement some virtual methods, it will compile but will also have some unresolved symbols.

See my comment on #15 for example. It took me 30 minutes to figure out why the plugin was not initialising, while with the message I would have found it on the go:
<pre>
[W] unknown:0 - cannot load plugin: "libsailfishsecrets-passwordagentauth.so" "Cannot load library /usr/lib/Sailfish/Secrets/libsailfishsecrets-passwordagentauth.so: (/usr/lib/Sailfish/Secrets/libsailfishsecrets-passwordagentauth.so: undefined symbol: _ZN8Sailfish7Secrets6Daemon7Plugins19PasswordAgentPlugin19beginAuthenticationEjx)"
</pre>

This MR is not a life changer neither correct a bug, but it makes life easier I guess when encountering small mistakes like that.

chriadam:
LGTM but can you please add the same also to the crypto plugin loading code, for symmetry?

vityafx:
I'd suggest not to copy-paste the code, but have a code-reuse in use: just put plugin loading in a common place and use this code from both places.

dcaliste:
@chriadam I've done the symmetric in crypto plugin handler.

@vityafx I've added a plugin helper to factorise the code from both secret and crypto handlers. Since it's private, I didn't put any namespace, if both of you think that it's better with, I'll update.

vityafx:
@chriadam can we merge this so I may use these changes for my `initialise` method?

chriadam:
LGTM thanks!

chriadam:
Program received signal SIGSEGV, Segmentation fault.
0x00000010 in ?? ()
(gdb) bt
#0  0x00000010 in ?? ()
#1  0x00045f8c in qobject_cast<Sailfish::Secrets::EncryptedStoragePlugin*> (object=object@entry=0xb26a8) at ../lib/Secrets/extensionplugins.h:300
#2  PluginHelper::storeAs<Sailfish::Secrets::EncryptedStoragePlugin> (this=this@entry=0xbefff2f8, obj=obj@entry=0xb26a8, store=store@entry=0x9e984, 
    category=category@entry=0x20399 <lcSailfishSecretsDaemon()>) at ./plugin_p.h:29
#3  0x00035c1e in Sailfish::Secrets::Daemon::ApiImpl::RequestProcessor::loadPlugins (this=this@entry=0x9e968, pluginDir=...)
    at SecretsImpl/secretsrequestprocessor.cpp:106
#4  0x00025ad2 in Sailfish::Secrets::Daemon::ApiImpl::SecretsRequestQueue::SecretsRequestQueue (this=0x936a8, parent=<optimized out>, pluginDir=..., 
    autotestMode=<optimized out>) at SecretsImpl/secrets.cpp:327
#5  0x0001ccea in Sailfish::Secrets::Daemon::Controller::Controller (this=0xbefff548, secretsPluginDir=..., cryptoPluginDir=..., 
    autotestMode=<optimized out>, parent=0x0) at controller.cpp:52
#6  0x0001c834 in main (argc=2, argv=<optimized out>) at main.cpp:38

PR #22, opened 2018-02-22
[sailfish-secrets] Add support for InteractionRequest. Contributes to JB#36797

chriadam:
Some client use cases require the capability to request user input
via a system-mediated UI where that input is NOT stored into the
secrets storage (for example, a daemon which wishes to use a transient
lock code for some functionality).

This commit adds support for an InteractionRequest where the user
is explicitly informed that the data they are entering should not
be considered "secure" in the secrets-storage sense.

Contributes to JB#36797


daemon/SecretsImpl/secrets.cpp
```diff
@@ -1107,14 +1133,14 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
...
+                        : QByteArray();
                 if (request->isSecretsCryptoRequest) {
-                    QByteArray userInput = request->outParams.size()
-                            ? request->outParams.takeFirst().value<QByteArray>()
-                            : QByteArray();
                     asynchronousCryptoRequestCompleted(request->cryptoRequestId, result, QVariantList() << userInput);
                 } else {
-                    // shouldn't happen!
-                    qCWarning(lcSailfishSecretsDaemon) << "Error: UserInputRequest not a secrets crypto request!";
+                    request->connection.send(request->message.createReply() << QVariant::fromValue<Result>(result)
```
vityafx: Interesting, it was "should not happen" earlier and now it is okay? Needs some explanation.

lib/Secrets/interactionrequest.cpp
```diff
@@ -0,0 +1,182 @@
...
+ * // Define the interaction request prompt parameters.
+ * Sailfish::Secrets::InteractionParameters uiParams;
+ * uiParams.setInputType(Sailfish::Secrets::InteractionParameters::AlphaNumericInput);
+ * uiParams.setEchoMode(Sailfish::Secrets::InteractionParameters::NormalEcho);
+ * uiParams.setPromptText(tr("Enter some data"));
+ *
+ * // Perform the request.
+ * Sailfish::Secrets::SecretManager sm;
+ * Sailfish::Secrets::InteractionRequest ir;
+ * ir.setManager(&sm);
```
dcaliste: It seems to me that ir.setInteractionParameters() is missing here.

lib/Secrets/interactionrequest.cpp
```diff
@@ -0,0 +1,182 @@
...
+InteractionParameters InteractionRequest::interactionParameters() const
+{
+    Q_D(const InteractionRequest);
+    return d->m_interactionParameters;
+}
+
+/*!
+ * \brief Sets the interaction parameters which allow customisation of the input prompt to \a uiParams
+ */
+void InteractionRequest::setInteractionParameters(const InteractionParameters &uiParams)
```
dcaliste: Why is it called set InteractionParameters() while it is called setUiParameters() in storesecretrequest?

daemon/SecretsImpl/secrets.cpp
chriadam: Because before, the only codepath the UserInputRequest was used in, was as an asynchronous crypto request (i.e. when getting input key data for a key derivation request).  Now it is used also for "standalone" secrets request (i.e. InteractionRequest).

lib/Secrets/interactionrequest.cpp
chriadam: Good catch, thanks will fix

lib/Secrets/interactionrequest.cpp
chriadam: Good point.  We should standardise.  I guess we should use setInteractionParameters() everywhere for consistency.  I will update those other cases.

Issue #23, opened 2018-02-26
Where to store and how to pass a user's password for Tokens?


Denis-Semakin:
Hello. In order to perform a common cryptographic operation using CryptoToki API (for Tokens) one need to call C_Login() function which requires a user's password. By default this password should be generated and set up on Token initialization process by security administrator. However user can change its own password.
So the question is how to pass this password to plugin's method since the parent abstract class CryptoPluginh has now such opportunity and parameters?

chriadam:
Hi!  I suspect we may need to add a "QVariantMap customParameters" argument to each method in the CryptoPlugin interface (similar to what was done for the KDF method).  This seems somewhat ugly, but probably necessary.  Do you agree, or do you think some other solution would be better?

Denis-Semakin:
@chriadam Hi. I think we don't need to pass "customParameters" in every single cryptoki method. User just need to be authorized to run them. Let me show a steps:
The following steps must be performed before any actual method will be called. So I suppose it should be invoked inside constructor of CryptoPlugin, if it's possible and then we need to pass (or to get it) a user's password there.
 - Initialize Cryptoki PKCS#11 library (IOW call dlopen());
 - GetSlotLits();
 - GetTokenInfo(); // To check out that library can interact with given token (FYI: several vendors can provide their proprietary library)
 - OpenSession() with token;
 - Login() // here all preparatory steps to work with token are ended. And this will be called only once.
Currently the functions that are enumerated above I've put in the constructor of the class LibLoader() which is called in CryptoTokiPligin constructor, understand that it's hard to listen this without seeing source code (I may push it in develop branch or etc BTW).
So then, after user is authorized one may perform any crypto functions: sign(), verify(), etc... without any password since session will be already opened.
Your opinion?

Denis-Semakin:
@chriadam 
A CryptoPlugin should provide a number of methods, e.g. sign(), verify(), decrypt(), etc...
To perform them with tokens a session with token have to be opened and user's password is used for this.
I see two ways to solve this:
 - pass password every time when user wants to call plugin;
 - open session with token only ones then use its handler.
So since the first way looks like incorrectly I've suggested to perform token initialization in constructor, although I understand this is not a good place to call it. That's why I was asking to your opinion.

Denis-Semakin:
@chriadam BTW I vote to your idea about initialize(parameters) method that you told on '#'OMP channel.
Then we need to design it correctly.

[...]

PR #28, opened 2018-03-02
Add initialise method to the extension plugins

vityafx:
Extension plugins may need some parameters for correct initialization. A simple example would be cryptotoki plugin which needs a password for working with the token, without it in can be accessed at all. So it was decided to provide a `QVariantMap` with all the parameters which might be needed by plugins. As of now, this map is empty because the data needed for us (master-lock password) is not implemented yet.


daemon/plugin_p.h
```diff
@@ -73,6 +74,8 @@ class PluginHelper : public QPluginLoader
             return 0;
         }
         qCDebug(category) << "loading plugin:" << fileName() << "with name:" << plugin->name();
+        qCDebug(category) << "initialising plugin:" << fileName();
+        plugin->initialise(parameters);
```
Denis-Semakin: Is this a place where real plugin (*.so file) is loaded?
If yes, then where is the result of initialize() check? it return true or false, then may be
```c
if (plugin->initialise(parameters))
...
else
...
```

lib/Crypto/extensionplugins.h
```diff
@@ -44,6 +44,8 @@ class SAILFISH_CRYPTO_API CryptoPlugin
     CryptoPlugin();
     virtual ~CryptoPlugin();
 
+    virtual bool initialise(const QVariantMap &parameters) = 0;
```
Denis-Semakin: initiali**z**e()

Denis-Semakin: @chriadam one guy told me that it's correctly in Australia? I mean 's'.

daemon/plugin_p.h
```diff
@@ -45,6 +45,7 @@ class PluginHelper : public QPluginLoader
 
     template <typename Plugin>
     Plugin* storeAs(QObject *obj, QMap<QString, Plugin*> *store,
+                    const QVariantMap &parameters,
```
dcaliste: Why not keep a const reference as a private member of PluginHelper and associate it at construction time? So like that we avoid to pass the same parameter argument at each storeAs() call. After all, parameter argument is the same for a single PluginHelper instance whatever we try to load it as. I would have like to do this for the category also but was not fluent enough in C++ to understand what should be the type of the private member…

vityafx: Sounds reasonable. Let me see if we can do that.

vityafx:
@dcaliste Review again please, I have moved it into the `PluginHelper` class itself.

daemon/plugin_p.h
```diff
@@ -85,6 +91,7 @@ class PluginHelper : public QPluginLoader
 private:
     FailureType m_failureType;
     bool m_autotestMode;
+    QVariantMap m_initialisationParameters;
```
dcaliste: Can ```m_initialisationParameters``` be declared ```const &``` so there is no copy at initialisation of the PluginHelper?

vityafx: If I do it so, there might be errors caused by use after free as we don't have lifetimes checked as in Rust.

dcaliste: Yes, I understand the problem of lifetime, but since PluginHelper is a small wrapper for local use it should be ok (but not guarantee by API). Well as you prefer, having parameters in the constructor is already fine, even with a copy. After all it's a useless optimisation not to copy, we won't have millions of parameters.

LGTM.

daemon/plugin_p.h
```diff
@@ -73,6 +73,12 @@ class PluginHelper : public QPluginLoader
             return 0;
         }
         qCDebug(category) << "loading plugin:" << fileName() << "with name:" << plugin->name();
+        qCDebug(category) << "initialising plugin:" << fileName();
```
chriadam: This second debug line is probably unnecessary

lib/Crypto/extensionplugins.h
chriadam: Yes, Australian English uses ise in such word-endings consistently.  British English is inconsistent between ise and ize, while American English consistently uses ize.  I will ask Joona which style we should prefer here, it may be that American English is preferred...

chriadam:
How should reinitialisation take place (e.g. when the user changes the master lock passphrase?)  I guess the plugins also need a "bool reinitialise(const QVariantMap &oldParams, const QVariantMap &newParams)" or similar (assuming that it is mandatory to provide the old and new passphrase)?  Or, should this just be another call to initialise() just with the new parameters/passphrase?

lib/Crypto/extensionplugins.h
Denis-Semakin: @chriadam Thanks for explanation. I think it's up to @vityafx

chriadam:
I've added lock/unlock(code)/setLockCode(old, new) API to the plugin interfaces in https://github.com/sailfishos/sailfish-secrets/pull/30 (and also call it appropriately when the master lock code is set by the client via a LockCodeRequest).

vityafx:
> How should reinitialisation take place (e.g. when the user changes the master lock passphrase?) I guess the plugins also need a "bool reinitialise(const QVariantMap &oldParams, const QVariantMap &newParams)" or similar (assuming that it is mandatory to provide the old and new passphrase)? Or, should this just be another call to initialise() just with the new parameters/passphrase?

There is no need in `oldParams` there. If the plugin needs to store the old data it will store it inside itself, let the user decide what to do with it, don't decide for the user. Also, there is no need in separate method `reinitialize`, we can just call the `initialize` instead.

> 

I've added lock/unlock(code)/setLockCode(old, new) API to the plugin interfaces in #30 (and also call it appropriately when the master lock code is set by the client via a LockCodeRequest).

Well. I'll try doing a rebase then and implement this fully.

vityafx:
I don't really understand, the #30 does the same so this PR is useless?
@chriadam

chriadam:
Regarding the oldParams - it may be that the plugin doesn't actually store the old key, instead it relies on the user providing that correct key.  So, in the case where the plugin is in locked state, with a previously set key being required to unlock it, then if the user wants to reinitialise the plugin (i.e. set a new lock code for it) the user should provide the old lock code plus the new lock code.  The plugin will unlock itself using the old lock code (e.g. decrypt some database) and then set the new lock lock (e.g. re-encrypt the database contents with the new lock code).

chriadam:
Regarding the PR 30: I think that one implemented most of what is required.  The only outstanding question in my mind is: do we need to add support for explicit plugin lock/unlock/setLockCode separately from the master lock?  See my comment: https://github.com/sailfishos/sailfish-secrets/issues/23#issuecomment-373591767 for more information there.

vityafx:
Okay, so the `initialise` method and whole this PR is useless then. We should use these 5 methods you provide in the #30.

PR #30, opened 2018-03-09
Masterlock

chriadam:


chriadam:
This is now ready for review.

lib/Crypto/extensionplugins.h
```diff
@@ -58,6 +58,12 @@ class SAILFISH_CRYPTO_API CryptoPlugin
     virtual QMap<Sailfish::Crypto::CryptoManager::Algorithm, QVector<Sailfish::Crypto::CryptoManager::KeyDerivationFunction> > supportedKeyDerivationFunctions() const = 0;
     virtual QMap<Sailfish::Crypto::CryptoManager::Algorithm, Sailfish::Crypto::CryptoManager::Operations> supportedOperations() const = 0;
 
+    virtual bool supportsLocking() const;
+    virtual bool isLocked();
```
Venemo: Shouldn't this also be `const`?

lib/Secrets/interactionparameters.h
```diff
@@ -59,28 +59,37 @@ class SAILFISH_SECRETS_API InteractionParameters {
...
-        Encrypt          = 128,
-        Decrypt          = 256,
-        Sign             = 512,
-        Verify           = 1024,
-        DeriveDigest     = 2048,
-        DeriveMac        = 4096,
-        DeriveKey        = 8192,
+        UnknownOperation    = 0,
+
+        RequestUserData     = 1<<0,   // usually used in conjunction with StoreSecret, i.e. store data requested from user.
```
Venemo: In other places, you use spaces next to the `<<` operator. For consistency, I suggest `1 << 0` here too.

lib/Secrets/extensionplugins.h
```diff
@@ -51,6 +51,12 @@ class SAILFISH_SECRETS_API EncryptionPlugin
     virtual Sailfish::Secrets::EncryptionPlugin::EncryptionType encryptionType() const = 0;
     virtual Sailfish::Secrets::EncryptionPlugin::EncryptionAlgorithm encryptionAlgorithm() const = 0;
 
+    virtual bool supportsLocking() const;
```
Venemo: The more I look at this, the more I think there should be a common base class for these plugin classes. Does not need to be part of this PR, but would definitely improve the code maintainability.

lib/Secrets/lockcoderequest.h
```diff
@@ -0,0 +1,85 @@
...
+public:
+    enum LockCodeRequestType {
+        ModifyLockCode = 0,
+        ProvideLockCode,
+        ForgetLockCode
+    };
+    Q_ENUM(LockCodeRequestType)
+
+    LockCodeRequest(QObject *parent = Q_NULLPTR);
+    ~LockCodeRequest();
```
Venemo: Should the destructor be `virtual` here? I would either suggest a `virtual` destructor, or marking the class as `final`.

lib/Secrets/lockcoderequest.h
```diff
@@ -0,0 +1,85 @@
...
+
+public:
+    enum LockCodeRequestType {
+        ModifyLockCode = 0,
+        ProvideLockCode,
+        ForgetLockCode
+    };
+    Q_ENUM(LockCodeRequestType)
+
+    LockCodeRequest(QObject *parent = Q_NULLPTR);
```
Venemo: Maybe mark the constructor with `explicit`.

lib/Secrets/extensionplugins.h
```diff
@@ -219,6 +237,12 @@ class SAILFISH_SECRETS_API AuthenticationPlugin : public QObject
     virtual Sailfish::Secrets::AuthenticationPlugin::AuthenticationTypes authenticationTypes() const = 0;
     virtual Sailfish::Secrets::InteractionParameters::InputTypes inputTypes() const = 0;
 
+    virtual bool supportsLocking() const;
+    virtual bool isLocked();
```
Venemo: I would suggest `const` for all `isLocked()` methods, I see there are a few of them, but I didn't comment on every one.

lib/Crypto/extensionplugins.h
chriadam: Done.  Previously it wasn't, because performing database select queries previously was non-const (due to preparation modifying the prepared query cache), however I've now made the cache mutable in order to make those methods const, and thus have now made the isLocked() methods const also.

lib/Secrets/interactionparameters.h
chriadam: Done

lib/Secrets/extensionplugins.h
chriadam: Right, let's do that in a separate PR

lib/Secrets/lockcoderequest.h
chriadam: All of the Request types derive from Request base class which has a virtual dtor.

lib/Secrets/extensionplugins.h
chriadam: Yep, done

lib/Secrets/lockcoderequest.h
chriadam: We should consider doing that, but in a separate PR as that would touch every eventual-QObject-derived class in the repository, and is not strictly related to this masterlock functionality.

blammit:
Looks good.

chriadam:
Thanks!  I assume OMP will have some feedback over the coming week or so, but I guess we can fix those issues later on, shouldn't hold up this PR from merging as I know that there are pending PRs which depend on this stuff.  Merging now.

PR #31, opened 2018-03-14
Jb40058 openssl features

blammit:


chriadam:
Minor note: JB#40058 tasks, probably best to use [sailfish-crypto] tag in commit messages, whereas for JB#36797 tasks use [sailfish-secrets].

plugins/opensslcryptoplugin/evp_p.h
```diff
@@ -8,6 +8,8 @@
 #ifndef SAILFISHCRYPTO_PLUGIN_CRYPTO_OPENSSL_EVP_P_H
 #define SAILFISHCRYPTO_PLUGIN_CRYPTO_OPENSSL_EVP_P_H
 
+#include "Crypto/cryptomanager.h"
```
chriadam: This is a C header, so probably just pass the block mode as an int to avoid having to pull in this header.
(Not that there's any need for this to be a C header, specifically, but it was copy-pasted from another C-only project.)

chriadam:
Mostly LGTM, thanks!

chriadam:
Also need unit tests etc

plugins/opensslcryptoplugin/evp_p.h
blammit: Yes true. Changing this means there will be hardcoded checks like (block_mode == 3), if that's what we want.

blammit:
Updated to fix issues and added unit tests.

These tests are failing at the moment regardless of whether my patches are applied. I can investigate this separately:
tst_cryptosecrets::secretsStoredKeys()
tst_cryptosecrets::cryptoStoredKey()
tst_cryptorequests::signVerify()
tst_cryptorequests::storedKeyRequests()
tst_cryptorequests::storedDerivedKeyRequests()
tst_cryptorequests::cipherEncryptDecrypt()

blammit:
These are failing when the patch is added, even with the default 256 key size. Needs investigation:
tst_cryptosecrets::secretsStoredKey()
tst_cryptosecrets::cryptoStoredKey()
tst_cryptorequests::storedDerivedKeyRequests()
tst_cryptorequests::storedGeneratedKeyRequests()
tst_cryptorequests::cipherEncryptDecrypt()

blammit:
Updated to rebase on top of the masterlock branch.

@chriadam I added the last two commits as we discussed, if you can check. The autotests now pass aside from the ones that were already failing against master.

plugins/opensslcryptoplugin/evp_p.h
Venemo: @chriadam Out of curiosity, where did you copy-paste it from?

blammit:
masterlock branch was merged. This is now rebased on top of master.

chriadam:
The changes to the input key handling (i.e. removing the hash-then-pad preamble) now causes breakages in the "custom lock code (from user)" codepaths (because the lock code is unlikely to be exactly the size required by the encryption algorithm.

I've created https://github.com/sailfishos/sailfish-secrets/pull/36 to address this issue, but there are a still a few failures which I ran out of time to fix, tonight.  I will continue on Monday.

chriadam:
I've now added a commit to derive the key from the authentication code properly in all cases (so that the cached key material is always a generated key, rather than sometimes a generated key and sometimes input key data (authentication code)) for the custom lock handling, etc.

blammit:
Yep, last commit looks good.

PR #32, opened 2018-03-14
Minor fixes

chriadam:


chriadam:
This is still WIP, need to finish the Secret implementation, then need to add the inline comparison operators for Key, then need to gadgetise and pimpl InteractionView.

After that's done, there are some doc fixes which I need to do (e.g. add the various request types which I've added recently), and I need to regenerate the DBus introspection data via qdbuscpp2xml.

Also, I will remove Certificate from the public API for now, since we won't have time to polish that one before the first release, I guess.

PR #36, opened 2018-03-16
Derive storage encryption key from authentication code

chriadam:


chriadam:
WIP, don't merge.

chriadam:
Not needed, I've pushed directly on top of the PR#31

PR #38, opened 2018-03-19
[sailfish-secrets] Allow plugin-target lock code request. Contributes to JB#36797

chriadam:
Previously, we only supported the bookkeeping database, a collection,
or a standalone secret to be the target of a code code request.
This commit adds support for specifying an extension plugin as the
target of the request, in case a particular plugin needs to be
explicitly unlocked via a non-masterlock (bookkeeping database key)
lock code request.

Contributes to JB#36797

### DIFF
--- a/daemon/CryptoImpl/crypto.cpp
+++ b/daemon/CryptoImpl/crypto.cpp
@@ -415,6 +415,60 @@ void Daemon::ApiImpl::CryptoDBusObject::finaliseCipherSession(
                                   result);
 }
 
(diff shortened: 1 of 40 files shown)


chriadam:
I am currently working on adding LockCodeRequest also to Crypto side (and routing it to Secrets side for user interaction flows etc).

chriadam:
Ok, this is now ready for review.

chriadam:
No negative comments from anyone so far... rebased and will now merge.

PR #39, opened 2018-03-19
Jb40058 openssl features blockmodes

blammit:


plugins/opensslcryptoplugin/cryptoplugin_common.cpp
```diff
@@ -101,6 +101,26 @@ namespace {
...
+        if (!initVector || blockMode == Sailfish::Crypto::CryptoManager::BlockModeEcb) {
+            // IV not required for this mode
+            return;
+        }
+
+        // Ensure the IV has the correct size. The IV size for most modes is the same as the block size.
+
+        switch (algorithm) {
+        case Sailfish::Crypto::CryptoManager::AlgorithmAes:
+            initVector->resize(16);  // AES = 128-bit block size
```
chriadam: In the case where the initialisation vector is too small, we should fill it with randomly generated bytes and emit the generated IV as per https://github.com/sailfishos/sailfish-secrets/blob/master/lib/Crypto/cipherrequest.cpp#L463.

In the case where the initialisation vector is too big, I guess we can truncate, but we still need to notify the client that the IV was "generated" / changed, so that they can send the "correct" IV to the recipient etc.

chriadam:
One comment.  We might have to add a "generatedInitialisationVector" property to the EncryptRequest also, to handle the case I noted.  Also, need to ensure that in the CipherRequest case we do emit that one as appropriate still (that one is unit tested in tst_cryptorequests::cipherEncryptDecrypt(), so it should be fine).

Aside from that, looks great.  I will merge https://github.com/sailfishos/sailfish-secrets/pull/40 first, so you will have to rebase on master once that goes in.

blammit:
Rebased against master against the #40 changes.

As discussed, the encryption/decryption in the crypto lib now fails if the IV is of the wrong size. We can add a feature later to automatically pad/crop the IV in this case.

chriadam:
LGTM.  Feel free to merge.  I noted support for GCM mode wasn't added yet - I added some boilerplate support for that mode in CipherRequest so it'd be nice if that were fully supported for that one, but doesn't have to be done right now, can wait if needed.  (See e.g. https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption)

PR #40, opened 2018-03-19
Various fixes 2nd attempt

Venemo:
Following are included:

* Compile on OpenSSL 1.1.0
* Fix plugin load mechanism to use `QT_PLUGIN_PATH`
* Remove `cryptoplugin_common.cpp`, instead use composition in `SqlCipherPlugin`


Venemo:
@chriadam Following up on your previous question: the new plugin load mechanism will not incur a significant performance penalty, as IIRC it only looks at the default Qt plugin path, where there aren't any actual plugins (just subdirectories). If this turns out to be not the case, we can improve further.

daemon/plugin_p.h
```diff
@@ -66,7 +66,7 @@ class PluginHelper : public QPluginLoader
             return 0;
         }
         if (plugin->name().endsWith(QStringLiteral(".test"), Qt::CaseInsensitive) != m_autotestMode) {
-            qCDebug(category) << "ignoring plugin:" << fileName() << "due to mode";
+            qCDebug(category) << "ignoring plugin:" << fileName() << "because we are in testing mode and it isn't a test plugin";
```
chriadam: The message is not correct, since we may be in non-test mode but the plugin is a test plugin, which would still cause this codepath to be hit.

database/database.cpp
```diff
@@ -315,6 +315,8 @@ bool Database::open(
 
     const QString databaseFile = databaseDir.absoluteFilePath(databaseFilename);
     const bool databasePreexisting = QFile::exists(databaseFile);
+    qCDebug(lcSailfishSecretsDaemonSqlite) << "databaseFile:" << databaseFile;
+    qCDebug(lcSailfishSecretsDaemonSqlite) << "databasePreexisting:" << databasePreexisting;
```
chriadam: Are these required?

chriadam:
I tested the plugin loading and you're correct, it does not attempt to descend into subdirs, so that's great!

However, this PR now causes the cipher unit test to fail:

FAIL!  : tst_cryptorequests::cipherEncryptDecrypt(128) Compared values are not the same
   Loc: [tst_cryptorequests.cpp(1253)]
FAIL!  : tst_cryptorequests::cipherEncryptDecrypt(192) Compared values are not the same
   Loc: [tst_cryptorequests.cpp(1184)]
FAIL!  : tst_cryptorequests::cipherEncryptDecrypt(256) Compared values are not the same
   Loc: [tst_cryptorequests.cpp(1184)]

chriadam:
I've updated the PR with fixes for the issues I noted, including updating the getFullKey() implementation to call the SqlCipherPlugin one when appropriate (which resolves the test failures).  This LGTM - if you're happy with it, feel free to merge :-)

PR #41, opened 2018-03-19
Sign and verify support for OpenSSL crypto plugin

Venemo:
Adds signing and verify support for the OpenSSL crypto plugin. Also contains some rudimentary tests which compare the output of our EVP code to the output of the OpenSSL command line utility. These are just some initial sanity tests, more tests will be added that will test the sign+verify in the plugin itself.


plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
```diff
@@ -417,15 +453,48 @@ Daemon::Plugins::OpenSslCryptoPlugin::verify(
...
+    }
+
+    // Verify the signature
+    r = osslevp_verify(evpDigestFunc, pkeyPtr, data.data(), data.length(), (const uint8_t*) signature.data(), (size_t) signature.length());
+    if (r == 1) {
+        // Verified successfully.
+        *verified = true;
+        return Sailfish::Crypto::Result(Sailfish::Crypto::Result::Succeeded);
+    } else if (r == 0) {
+        // Verification performed without error, but was unsuccessful.
```
chriadam: I think this should return Result::Succeeded but set *verified = false.

chriadam:
Looking good!  Still needs test cases added to void tst_cryptorequests::signVerify(), also does signing work with symmetric keys (e.g. a keyed hash signature)?

Venemo:
@chriadam "In theory" it should work with whatever keys can be loaded into `EVP_PKEY` by `PEM_read_bio_PrivateKey`. So far, I've only tested with the RSA thing that you can see in my initial tests.

Venemo:
I also added an implementation for `OpenSslCryptoPlugin::calculateDigest` (with `PaddingNone`) and one very simple test case.

Remaining to do in this branch:
* Support for padding (how important is that?)
* More tests for digest + sign + verify
* MAC support, possibly needs new method with arguments of type `CryptoManager::MessageAuthenticationCode` + `CryptoManager::Algorithm`

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
```diff
@@ -391,13 +391,35 @@ Daemon::Plugins::OpenSslCryptoPlugin::calculateDigest(
...
+    // Get the EVP digest function
+    const EVP_MD *evpDigestFunc = getEvpDigestFunction(digestFunction);
+    if (!evpDigestFunc) {
+        return Sailfish::Crypto::Result(Sailfish::Crypto::Result::UnsupportedDigest,
+                                        QLatin1String("Unsupported digest function chosen."));
+    }
+
+    // Create digest
+    uint8_t *digestBytes;
+    size_t digestLength;
```
chriadam: initialise local variables to zero

plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -285,3 +285,261 @@ int osslevp_aes_decrypt_ciphertext(int block_mode,
...
+    * bytes: data to digest
+    * bytesCount: the number of bytes in 'bytes'
+    * digest: where the generated digest will be stored, which will have to be freed using OPENSSL_free
+    * digestLength: where the length of the generated digest will be stored
+    Return value:
+    * 1 when the operation was successful.
+    * less than 0 when there was an error.
+ */
+int osslevp_digest(const EVP_MD *digestFunc,
+                 const void *bytes,
```
Denis-Semakin: Alignment? Missed space?

plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -285,3 +285,261 @@ int osslevp_aes_decrypt_ciphertext(int block_mode,
...
+                "failed to finalize DigestSign (2nd call)");
+        return -1;
+    }
+
+    // Set correct length to the output argument
+    *digestLength = actualDigestLength;
+
+    EVP_MD_CTX_destroy(mdctx);
+    return 1;
+}
```
Denis-Semakin: Don't like **goto** operator?
```c
if (some_error) {
   printf("Err");
   goto err_label;
}
...
err_label:
   free_resourses(...)
```
Kernel Coding Style

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
Venemo: Will do.

plugins/opensslcryptoplugin/evp/evp.c
Venemo: @Denis-Semakin Bad copy-paste. Will fix.

plugins/opensslcryptoplugin/evp/evp.c
Venemo: @Denis-Semakin This is how the code looked like, so I just followed the existing style.
@chriadam Are we okay with `goto`? I guess it would be handy for error handling, indeed.

Denis-Semakin: @Venemo BTW for some resources, e.g. opened file descriptors, pointers to allocated memory chunks that need to be freed... one may use **__attribute__((cleanup))** expression.

Venemo: @Denis-Semakin To be honest, I would personally prefer to just use C++ smart pointers, but it's a C file, and I've been too afraid to ask why it needs to be a C file. But I agree that this sort of duplicated code must be cleaned up one way or the other. I'd be fine with either.

Venemo:
Thanks for the reviews!  
I now rebased on top of current master, and fixed some stuff that you guys suggested.

Venemo:
Note to self:  
Perhaps this can be used with padding: https://www.openssl.org/docs/man1.1.0/crypto/EVP_PKEY_CTX_set_rsa_padding.html ― but it seems not every padding type is supported with every algorithm. In particular, the name `EVP_PKEY_CTX_set_rsa_padding` suggests that it can only do this in case of RSA. So I would guess `EVP_PKEY_get0_RSA` can be used to find out if it is an RSA key, and if not, we can return `UnsupportedOperation`.

Venemo:
Sorry, GitHub messed up something after I rebased... will close this and open a new PR instead.

PR #42, opened 2018-03-21
[sailfish-crypto] Add autotests for GCM mode (for AES encryption). Contributes to JB#40058

blammit:
WIP. The test fails at the moment because the decrypted plaintext has an extra 16 bytes on the end. Looking at the plugin, it looks like GCM decryption is implemented correctly as per the openssl docs, so this needs some investigation as to why it's failing.


chriadam:
The data emitted during finalisation of encryption stage is not ciphertext in GCM mode, but tag data.  I've added an if-check for the mode there, and now everything works.  Thanks, merging!

PR #43, opened 2018-03-21
Sign, verify and digest support for OpenSSL crypto plugin

Venemo:
Adds support for:
* Signing with RSA private key
* Verifying a signature with an RSA public key
* Calculating a digest

Tests:
* end-to-end tests in `tst_cryptorequests` for signing, verify and digest
* openssl command line tests to ensure that our EVP code is sane

Current limitations:
* only supports `PaddingNone`
* no support for MAC (message authentication codes)

I believe this is now ready to be merged.


Venemo:
GitHub screwed up this PR too, after I did an intelligent rebase. Sorry for the mess, will open another PR

PR #44, opened 2018-03-21
Sign, verify and digest support for OpenSSL crypto plugin

Venemo:
Adds support for:
* Signing with RSA private key
* Verifying a signature with an RSA public key
* Calculating a digest

Tests:
* end-to-end tests in `tst_cryptorequests` for signing, verify and digest
* openssl command line tests to ensure that our EVP code is sane

Current limitations:
* only supports `PaddingNone`
* no support for MAC (message authentication codes)

I believe this is now ready to be merged.

### DIFF
--- a/tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
+++ b/tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
@@ -12,7 +12,9 @@
 #include <QObject>
 #include <QElapsedTimer>
 #include <QFile>
+#include <QtCore/QCryptographicHash>
 
+#include "Crypto/calculatedigestrequest.h"
 #include "Crypto/cipherrequest.h"
 #include "Crypto/decryptrequest.h"
 #include "Crypto/deletestoredkeyrequest.h"
@@ -90,6 +92,7 @@ private slots:
     void generateKeyEncryptDecrypt();
     void validateCertificateChain();
     void signVerify();
+    void calculateDigest();
     void storedKeyRequests_data();
     void storedKeyRequests();
     void storedDerivedKeyRequests_data();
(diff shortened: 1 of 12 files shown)


tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -429,12 +484,45 @@ void tst_cryptorequests::signVerify()
...
+
+    WAIT_FOR_FINISHED_WITHOUT_BLOCKING(cdr);
+    QCOMPARE(cdrss.count(), 2);
+    QCOMPARE(cdr.status(), Request::Finished);
+
+    QCOMPARE(cdr.result().code(), Result::Succeeded);
+    QCOMPARE(cdrds.count(), 1);
+
+    QByteArray digest = cdr.digest();
+    QVERIFY2(digest.length() != 0, "Calculated digest should NOT be empty.");
```
chriadam: You could also check that the digest value is the same as produced by QCryptographicHash via Sha256, to ensure that the digest is calculated "correctly".

chriadam:
In my Mer SDK build env, I get:

In file included from ../../../plugins/opensslcryptoplugin/evp/evp.c:8:0:
../../../plugins/opensslcryptoplugin/evp/evp_p.h:52:18: error: unknown type name 'uint8_t'
                  uint8_t **digest,

chriadam:
Regarding why evp_p.h was a C-compatible header, it's because I originally wrote it for a (BSD-licensed) C library, and when I copied into this project I didn't bother to convert it to C++.  It should be converted to C++, or alternatively using kernel style in it is also fine.

chriadam:
Need to package tst_evp for installation too.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -386,18 +388,59 @@ void tst_cryptorequests::validateCertificateChain()
...
 void tst_cryptorequests::signVerify()
 {
-    // TODO: sign/verify not yet implemented in test plugin.
-    QByteArray plaintext = "Test plaintext data", signature;
+    // Generate RSA key for signing
+    // ----------------------------
+
+    // Create key template
+    Key keyTemplate;
+    keyTemplate.setSize(1024);
```
chriadam: for RSA keys, the key size is set via the RsaKeyPairGenerationParameters (modulusLength)

chriadam:
I've fixed the couple of small issues I found.  All the tests now pass on device.

Looks great, I'll merge now.  Going forward, we should improve the plugin by supporting both DSA and ECDSA for sign/verify functions, and also supporting more digest functions (currently just EVP_sha256() is supported from what I can tell).

PR #45, opened 2018-03-23
Add CTR and XTS block support for AES

blammit:


plugins/opensslcryptoplugin/evp/evp_helpers_p.h
```diff
@@ -144,3 +144,93 @@ const EVP_MD *getEvpDigestFunction(Sailfish::Crypto::CryptoManager::DigestFuncti
     }
 }
 
+const EVP_CIPHER *getEvpCipher(int block_mode, int key_length_bytes)
```
Venemo: This is exactly why I introduced `evp_helpers_p.h` so that we can put the glue between sailfish-crypto and evp here. Good stuff!

PR #46, opened 2018-03-23
EC key generation, new digest functions, RSA encryption

Venemo:
Implements EC key generation and adds new digest functions.

### DIFF
--- a/lib/Crypto/keypairgenerationparameters.cpp
+++ b/lib/Crypto/keypairgenerationparameters.cpp
@@ -219,6 +219,7 @@ EcKeyPairGenerationParameters::EcKeyPairGenerationParameters()
     : KeyPairGenerationParameters()
 {
     setKeyPairType(KeyPairGenerationParameters::KeyPairEc);
+    setEllipticCurve(CryptoManager::CurveSecp384r1);
 }
 
 /*!
--- a/plugins/opensslcryptoplugin/evp/evp.c
+++ b/plugins/opensslcryptoplugin/evp/evp.c
@@ -254,6 +254,129 @@ int osslevp_aes_decrypt_ciphertext(const EVP_CIPHER *evp_cipher,
 }
 
 /*
+    int osslevp_pkey_encrypt_plaintext(EVP_PKEY *pkey,
+                                       const unsigned char *plaintext,
+                                       size_t plaintext_length,
+                                       uint8_t **encrypted,
+                                       size_t *encrypted_length);
+
+    Implements encryption with an asymmetric key.
+    See: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_encrypt.html
+
+    Arguments:
+    * pkey: key used for encryption
+    * plaintext: the data to encrypt
+    * plaintext_length: number of bytes in 'plaintext'
+    * encrypted: output argument, where memory will be allocated for the
+    *            encrypted data, needs to be freed with OPENSSL_free
+    * encrypted_length: output argument, where the number of encrypted
+    *                   bytes will be put.
+
+    Return value:
+    * 1 when the operation was successful
+    * less than 0 when there was an error
+ */
+int osslevp_pkey_encrypt_plaintext(EVP_PKEY *pkey,
+                                   const uint8_t *plaintext,
+                                   size_t plaintext_length,
+                                   uint8_t **encrypted,
+                                   size_t *encrypted_length)
+{
+    int r = -1;
+
+    EVP_PKEY_CTX *pkctx = EVP_PKEY_CTX_new(pkey, NULL);
+    OSSLEVP_HANDLE_ERR(pkctx == NULL, r = -1, "failed to create EVP_PKEY_CTX", err_dontfree);
+
+    r = EVP_PKEY_encrypt_init(pkctx);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to initialize EVP_PKEY_CTX for encryption", err_free_pkctx);
+
+    // TODO: padding support
+    //r = EVP_PKEY_CTX_set_rsa_padding(pkctx, RSA_NO_PADDING);
+    //OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to set RSA padding", err_free_pkctx);
+
+    r = EVP_PKEY_encrypt(pkctx, NULL, encrypted_length, plaintext, plaintext_length);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to calculate PKEY encrypted size", err_free_pkctx);
+
+    *encrypted = OPENSSL_malloc(*encrypted_length);
+    OSSLEVP_HANDLE_ERR(*encrypted == NULL, r = -1, "failed to allocate memory for encrypted data", err_free_pkctx);
+
+    r = EVP_PKEY_encrypt(pkctx, *encrypted, encrypted_length, plaintext, plaintext_length);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to encrypt with PKEY", err_free_encrypted);
+
+    r = 1;
+    goto success;
+
+    err_free_encrypted:
+    OPENSSL_free(*encrypted);
+    success:
+    err_free_pkctx:
+    EVP_PKEY_CTX_free(pkctx);
+    err_dontfree:
+    return r;
+}
+
+/*
+    int osslevp_pkey_decrypt_ciphertext(EVP_PKEY *pkey,
+                                        const unsigned char *ciphertext,
+                                        size_t ciphertext_length,
+                                        uint8_t **decrypted,
+                                        size_t *decrypted_length);
+
+    Decrypts the given ciphertext using the supplied key.
+
+    Arguments:
+    * pkey: key used for decryption
+    * ciphertext: the data to decrypt
+    * ciphertext_length: number of bytes in 'ciphertext'
+    * decrypted: output argument, where memory will be allocated for the
+    *            decrypted data, needs to be freed with OPENSSL_free
+    * decrypted_length: output argument, where the number of decrypted
+    *                   bytes will be put.
+
+    Return value:
+    * 1 when the operation was successful
+    * less than 0 when there was an error
+*/
+int osslevp_pkey_decrypt_ciphertext(EVP_PKEY *pkey,
+                                    const unsigned char *ciphertext,
+                                    size_t ciphertext_length,
+                                    uint8_t **decrypted,
+                                    size_t *decrypted_length)
+{
+    int r = -1;
+
+    EVP_PKEY_CTX *pkctx = EVP_PKEY_CTX_new(pkey, NULL);
+    OSSLEVP_HANDLE_ERR(pkctx == NULL, r = -1, "failed to create EVP_PKEY_CTX", err_dontfree);
+
+    r = EVP_PKEY_decrypt_init(pkctx);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to initialize EVP_PKEY_CTX for decryption", err_free_pkctx);
+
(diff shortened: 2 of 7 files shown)


plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -7,6 +7,17 @@
 
 #include "evp_p.h"
 
+#define OSSLEVP_PRINT_ERR(message) \
+    fprintf(stderr, "%s#%d, %s: %s\n", __FILE__, __LINE__, __FUNCTION__, message);
```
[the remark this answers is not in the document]
Venemo: That is actually not new in this commit, that's how it was previously as well, ie. we print possible errors to `stderr`. Feel free to chime in if you have a better idea.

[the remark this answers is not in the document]
monich: How about returning the error to the caller and letting the caller print the error message in the manner the caller finds appropiate? There are Qt logging facilities, glib etc. Plenty of ways to make debugging and error messages available on demand or return them to the caller if the caller is interested. But that's what I like, and it could be that some people like to see those printed unconditionally to `stderr`, why not. Feel free to ignore. I just thought it was an oversight but apparently it's not.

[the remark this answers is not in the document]
Venemo: If you look at the calling code (`OpenSslCryptoPlugin`), you can see that an error is actually returned to the caller. However, since this is a daemon which is mostly used through D-Bus, the client cannot be expected to deal with low-level error codes from OpenSSL.

I guess `stderr` is a cheap way for debugging such issues. However this discussion is not really related to this pull request, since `stderr` was already used before. So I would suggest to open an issue where you can detail your suggestion and possible alternatives and we can continue the discussion.

@chriadam What do you think? Should we stick to `stderr` or use something else?

lib/Crypto/keypairgenerationparameters.cpp
```diff
@@ -219,6 +219,8 @@ EcKeyPairGenerationParameters::EcKeyPairGenerationParameters()
     : KeyPairGenerationParameters()
 {
     setKeyPairType(KeyPairGenerationParameters::KeyPairEc);
+    // TODO: is this a good default choice?
+    setEllipticCurve(CryptoManager::CurveSecp521r1);
```
chriadam: I'm not sure how many web-browsers etc support the NIST-P521 curve (aka SECP521r1), perhaps NIST-P384 or NIST-P256 would be better default choices.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -108,7 +113,6 @@ private slots:
     void addCryptoTestData()
     {
         QTest::addColumn<CryptoManager::BlockMode>("blockMode");
-        QTest::addColumn<int>("keySize");
```
chriadam: was this removed intentionally?

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -392,29 +396,55 @@ void tst_cryptorequests::validateCertificateChain()
...
+static inline KeyPairGenerationParameters *getKeyPairGenerationParameters(CryptoManager::Algorithm algorithm)
+{
+    switch (algorithm)
+    {
+    case CryptoManager::AlgorithmRsa:
+        return new RsaKeyPairGenerationParameters();
+    case CryptoManager::AlgorithmEc:
+        return new EcKeyPairGenerationParameters();
+    default:
+        throw std::runtime_error("Unknown algorithm");
```
chriadam: instead of throwing, maybe just return a default constructed KeyPairGenerationParameters() and QSKIP in that case.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -392,29 +396,55 @@ void tst_cryptorequests::validateCertificateChain()
...
+    QTest::newRow("EC + SHA256") << CryptoManager::AlgorithmEc << CryptoManager::DigestSha256;
+    QTest::newRow("EC + SHA512") << CryptoManager::AlgorithmEc << CryptoManager::DigestSha512;
+}
+
+static inline KeyPairGenerationParameters *getKeyPairGenerationParameters(CryptoManager::Algorithm algorithm)
+{
+    switch (algorithm)
+    {
+    case CryptoManager::AlgorithmRsa:
+        return new RsaKeyPairGenerationParameters();
```
chriadam: why construct on the heap?

plugins/opensslcryptoplugin/evp/evp.c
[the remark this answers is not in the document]
chriadam: It's a good idea to try to return the "actual" error to the client for debugging purposes.  By which I mean, it's not really (IMO) useful to try to modify application behaviour at runtime depending on the error message, but it is useful for client applications to see what they might have done wrongly...

In short, I think it's probably a good idea, but very low priority.

chriadam:
A couple of minor issues, but mostly looking great.  Thanks!

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
Venemo: Not sure, maybe a bad merge.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
Venemo: I was under the impression that the extra data from the derived classes would be lost if I just passed this on the stack.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
Venemo: Ok!

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
chriadam: It shouldn't be - and if it is, that's a bug we need to fix.  The classes should be immune to slicing etc.

chriadam:
What's the ETA on this PR being updated?

Venemo:
@chriadam Doing it today. Sorry for the delay.

Venemo:
Rebased on top of current master and fixed the issues raised in the reviews.

Venemo:
Added the ability to encrypt and decrypt with RSA.

Venemo:
@blammit Can you please review these changes?

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -109,42 +114,60 @@ private slots:
 private:
     void addCryptoTestData()
     {
+        QTest::addColumn<CryptoManager::Algorithm>("algorithm");
```
blammit: Note that cipherEncryptDecrypt_data() needs to have the CryptoManager::AlgorithmAes test parameter added to the test data rows as well.

plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -240,6 +253,129 @@ int osslevp_aes_decrypt_ciphertext(const EVP_CIPHER *evp_cipher,
...
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to calculate PKEY encrypted size", err_free_pkctx);
+
+    *encrypted = OPENSSL_malloc(*encrypted_length);
+    OSSLEVP_HANDLE_ERR(*encrypted == NULL, r = -1, "failed to allocate memory for encrypted data", err_free_pkctx);
+
+    r = EVP_PKEY_encrypt(pkctx, *encrypted, encrypted_length, plaintext, plaintext_length);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to encrypt with PKEY", err_free_encrypted);
+
+    r = 1;
+    goto success;
```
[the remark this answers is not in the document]
Venemo: We have agreed with @chriadam to follow this style, to reduce code duplication (ie. repeating all the `free` calls everywhere).

plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -240,6 +253,129 @@ int osslevp_aes_decrypt_ciphertext(const EVP_CIPHER *evp_cipher,
...
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to initialize EVP_PKEY_CTX for decryption", err_free_pkctx);
+
+    // TODO: padding support
+    //r = EVP_PKEY_CTX_set_rsa_padding(pkctx, RSA_NO_PADDING);
+    //OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to set RSA padding", err_free_pkctx);
+
+    r = EVP_PKEY_decrypt(pkctx, NULL, decrypted_length, ciphertext, ciphertext_length);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "failed to calculate PKEY encrypted size", err_free_pkctx);
+
+    *decrypted = (uint8_t*) OPENSSL_malloc(*decrypted_length * 2);
```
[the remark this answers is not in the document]
Venemo: This is a mistake. Thanks for noticing.

plugins/opensslcryptoplugin/opensslcryptoplugin.h
```diff
@@ -174,6 +174,50 @@ class OPENSSLCRYPTOPLUGIN_EXPORT OpenSslCryptoPlugin : public QObject, public Sa
...
+
+    Sailfish::Crypto::Result encryptAes(
+            const QByteArray &data,
+            const QByteArray &iv,
+            const Sailfish::Crypto::Key &key,
+            Sailfish::Crypto::CryptoManager::BlockMode blockMode,
+            Sailfish::Crypto::CryptoManager::EncryptionPadding padding,
+            QByteArray *encrypted);
+
+    Sailfish::Crypto::Result encryptAssymmetric(
```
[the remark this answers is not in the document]
Venemo: Will fix, thanks.

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
```diff
@@ -612,15 +731,123 @@ Daemon::Plugins::OpenSslCryptoPlugin::encrypt(
...
+                                        QLatin1String("TODO: initialization vectors are not yet supported with asymmetric encryption"));
+    }
+
+    if (fullKey.algorithm() < Sailfish::Crypto::CryptoManager::FirstAsymmetricAlgorithm
+           || fullKey.algorithm() > Sailfish::Crypto::CryptoManager::LastAsymmetricAlgorithm) {
+        return Sailfish::Crypto::Result(Sailfish::Crypto::Result::UnsupportedOperation,
+                                        QLatin1String("OpenSslCryptoPlugin::encryptAssymmetric only supports asymmetric algorithms"));
+    }
+
+    if (padding != Sailfish::Crypto::CryptoManager::EncryptionPaddingNone) {
```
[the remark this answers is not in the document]
Venemo: That's just bad copy-paste. Will fix, thanks.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
Venemo: Will fix, thanks!

plugins/opensslcryptoplugin/evp/evp.c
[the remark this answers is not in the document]
blammit: Okay, sounds fine if you've discussed it then.

Venemo:
Fixed all issues raised by @blammit ― will merge later today unless I get any more comments

Venemo:
Now also added support for some RSA padding schemes.

Issue #35, opened 2018-03-15
Reduce code duplication


Venemo:
I'm just writing it here in order not to forget.  
I noticed there are several places where code is duplicated. So here are my current thoughts, I would welcome some feedback from @blammit and @chriadam on them. We would need to fix these eventually if we don't want it to become unmaintainable.

* `evp_p.h` (will be fixed by Bea's PR)
* plugin loading mechanism is needlessly duplicated between secrets and crypto
* some basic methods and enums are present in both secrets and crypto under different namespaces
* in the masterlock branch the plugin locking is duplicated between different plugin types

Possible solutions:

1. Make either crypto depend on secrets, or secrets on crypto. Put all basic declarations in one, and use those in the other.
2. Create a common base for both secrets and crypto. Put all basic declarations there.
3. Reduce the functional overlapping between the two somehow.

What do you guys think?

blammit:
Yes it would be ideal to rework things to greatly reduce the duplication. My understanding is that secrets is about storage and therefore it would depend on crypto rather than the other way around; so, is it a problem if secrets has a hard dependency on crypto? If not, then take that path, and otherwise make a common library that both can use. @chriadam would have a clearer overview of this.

chriadam:
I do agree that the various plugin types should probably extend from a common base class where possible, and that the openssl code duplication should be minimised (the _common.cpp file already does this to a large extent, but there is more to do, e.g. the evp.h as you pointed out already).

In regards to the other points raised:

1) I would greatly prefer to avoid any build time requirement between Secrets and Crypto client API libraries.  There are many use-cases for Secrets which don't require Crypto, and vice-versa.

2) Common base for "basic declarations" would probably require either polluting Sailfish namespace, or the global namespace, either of which seem like suboptimal options, especially since the type-names (or enum names) which might be shared are likely candidates for collisions in the future.  First step might be to see what basic declarations are in fact used across both - I suspect it's a very small number of things.

3) Some functional overlap exists (specifically, Keys can be stored securely, and as an implementation detail these are stored as Secrets), but cannot really be reduced.

chriadam:
While it is definitely true that to be maximally secure, a plugin will most likely want to implement both EncryptedStorage and Crypto plugin interfaces, that is unrelated to the client application requirements.  

[...]

PR #47, opened 2018-03-26
[sailfish-secrets] Serialise plugin operations via QtConcurrent. Contributes to JB#36797

chriadam:
We define two thread pools, each of which contains a single thread,
which are used to run functions exposed by plugins of the Secrets
and Crypto domains respectively.

This ensures:
1) the main thread will not be blocked synchronously executing
   long cryptographic operations provided by client plugins
2) client plugins don't have to manage operation serialisation
   as they are guaranteed to be serialised and non-interfering

This commit further then ensures that the secrets I/O thread pool
is used to perform plugin operations for secrets API requests.

Contributes to JB#36797


chriadam:
WIP!  Still need to implement the Crypto side.  (Note that CryptoStorage plugins need to run on the Secrets thread!)

daemon/SecretsImpl/pluginfunctionwrappers_p.h
```diff
@@ -0,0 +1,296 @@
...
+namespace Sailfish {
+
+namespace Secrets {
+
+namespace Daemon {
+
+namespace ApiImpl {
+
+std::tuple<bool, // found
+           Sailfish::Secrets::Result> lockSpecificPlugin(
```
Venemo: Suggestion:

```cpp
using LockResult = std::tuple<bool, Sailfish::Secrets::Result>;
```

daemon/SecretsImpl/pluginfunctionwrappers_p.h
```diff
@@ -0,0 +1,296 @@
...
+
+namespace EncryptionPluginWrapper {
+    bool isLocked(Sailfish::Secrets::EncryptionPlugin *plugin);
+    bool lock(Sailfish::Secrets::EncryptionPlugin *plugin);
+    bool unlock(Sailfish::Secrets::EncryptionPlugin *plugin,
+                const QByteArray &lockCode);
+    bool setLockCode(Sailfish::Secrets::EncryptionPlugin *plugin,
+                     const QByteArray &oldLockCode,
+                     const QByteArray &newLockCode);
+    std::tuple<Sailfish::Secrets::Result, QByteArray> deriveKeyFromCode(
```
Venemo: This contradicts the current sailfish-secrets style, where we pass additional output arguments as pointers. I would suggest to change the signature to return a `Result` (like everything else), and add the `QByteArray` as a pointer argument.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -1814,14 +1822,60 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
...
+                    ? request->outParams.takeFirst().value<Result>()
+                    : Result(Result::UnknownError,
+                             QLatin1String("Unable to determine result of FindCollectionSecretsRequest request"));
+            if (result.code() == Result::Pending) {
+                // shouldn't happen!
+                qCWarning(lcSailfishSecretsDaemon) << "FindCollectionSecretsRequest:" << request->requestId << "finished as pending!";
+                *completed = true;
+            } else {
+                QVector<Secret::Identifier> identifiers = request->outParams.size()
+                        ? request->outParams.takeFirst().value<QVector<Secret::Identifier> >()
```
Venemo: Since C++11, the compiler can distinguish between the `>>` operator and template syntax, so you don't need to use `> >` here. Suggestion: `.value<QVector<Secret::Identifier>>()` (without the extra space between the two `>`s)

daemon/SecretsImpl/secrets.cpp
```diff
@@ -1814,14 +1822,60 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
...
+                        : QVector<Secret::Identifier>();
+                if (request->isSecretsCryptoRequest) {
+                    asynchronousCryptoRequestCompleted(request->cryptoRequestId, result, QVariantList() << QVariant::fromValue<QVector<Secret::Identifier> >(identifiers));
```
Venemo: This style makes the code hard to read. Suggestion:
* use C++11 initializer lists instead
* also, the compiler can figure out the template argument on its own in this case, no need to explicitly write it.

```cpp
asynchronousCryptoRequestCompleted(request->cryptoRequestId, result, { QVariant::fromValue(identifiers) });
```

daemon/SecretsImpl/secrets.cpp
```diff
@@ -1814,14 +1822,60 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
...
+                } else {
+                    request->connection.send(request->message.createReply() << QVariant::fromValue<Result>(result)
+                                                                            << QVariant::fromValue<QVector<Secret::Identifier> >(identifiers));
```
Venemo: This coding style is hard to follow.  
Suggestion:

```cpp
auto reply = request->message.createReply();
reply << QVariant::fromValue(result);
reply << QVariant::fromValue(identifiers);
request->connection.send(reply);
```

This is equivalent, but more readable.

daemon/SecretsImpl/secrets.cpp
Venemo: Same goes for every similar case.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -1814,14 +1822,60 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
...
+                }
+                *completed = true;
+            }
+            break;
+        }
+        case FindStandaloneSecretsRequest: {
+            Result result = request->outParams.size()
+                    ? request->outParams.takeFirst().value<Result>()
+                    : Result(Result::UnknownError,
+                             QLatin1String("Unable to determine result of FindStandaloneSecretsRequest request"));
```
Venemo: Perhaps create a small re-usable helper function for this pattern. It is repeated a lot.

```cpp
template <typename T>
T takeFirstOrDefault(QVariantList &list, const T &defaultValue) {
    if (list.size()) {
        return list.takeFirst().value<T>();
    }
    return defaultValue;
}
```

Or something similar.  
In fact, I'm surprised to see that `QList` doesn't have a built-in method for this.

daemon/SecretsImpl/pluginfunctionwrappers_p.h
Venemo: This would reduce a lot of repetition.

daemon/CryptoImpl/crypto.cpp
```diff
@@ -497,6 +499,11 @@ Daemon::ApiImpl::CryptoRequestQueue::~CryptoRequestQueue()
 {
 }
 
+QThreadPool* Daemon::ApiImpl::CryptoRequestQueue::cryptoThreadPool()
```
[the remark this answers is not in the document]
chriadam: The thread pool is owned by the request queue, and the request queue is the only thing which schedules requests to be run.

daemon/CryptoImpl/crypto.cpp
```diff
@@ -497,6 +499,11 @@ Daemon::ApiImpl::CryptoRequestQueue::~CryptoRequestQueue()
...
+{
+    return &m_cryptoThreadPool;
```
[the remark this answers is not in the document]
chriadam: Fair enough.  I will do so.

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,973 @@
...
+
+/* These methods are to be called via QtConcurrent */
+
+std::tuple<bool, Result> Daemon::ApiImpl::lockSpecificPlugin(
+        const QMap<QString, StoragePlugin*> &storagePlugins,
+        const QMap<QString, EncryptionPlugin*> &encryptionPlugins,
+        const QMap<QString, EncryptedStoragePlugin*> &encryptedStoragePlugins,
+        const QString &lockCodeTarget)
+{
+    if (storagePlugins.contains(lockCodeTarget)) {
```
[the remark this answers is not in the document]
chriadam: Yes, Venemo and I have discussed making a base class for the extension plugins, in order to do such refactoring.  However, it's low priority at the moment, so we'll do it later.

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,973 @@
...
+}
+
+std::tuple<bool, Result> Daemon::ApiImpl::modifyLockSpecificPlugin(
+        const QMap<QString, StoragePlugin*> &storagePlugins,
+        const QMap<QString, EncryptionPlugin*> &encryptionPlugins,
+        const QMap<QString, EncryptedStoragePlugin*> &encryptedStoragePlugins,
+        const QString &lockCodeTarget,
+        std::tuple<QByteArray, QByteArray> newAndOldLockCode)
+{
+    if (storagePlugins.contains(lockCodeTarget)) {
```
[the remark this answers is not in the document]
chriadam: Ditto

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,973 @@
...
+    return std::make_tuple(false, Result(Result::Failed));
+}
+
+bool Daemon::ApiImpl::lockPlugins(
+        QList<EncryptionPlugin*> encryptionPlugins,
+        QList<StoragePlugin*> storagePlugins,
+        QList<EncryptedStoragePlugin*> encryptedStoragePlugins)
+{
+    bool allSucceeded = true;
+    for (EncryptionPlugin *eplugin : encryptionPlugins) {
```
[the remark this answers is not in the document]
chriadam: Ditto

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,973 @@
...
+}
+
+bool Daemon::ApiImpl::unlockPlugins(
+        QList<EncryptionPlugin*> encryptionPlugins,
+        QList<StoragePlugin*> storagePlugins,
+        QList<EncryptedStoragePlugin*> encryptedStoragePlugins,
+        const QByteArray &encryptionKey)
+{
+    bool allSucceeded = true;
+    for (EncryptionPlugin *eplugin : encryptionPlugins) {
```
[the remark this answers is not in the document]
chriadam: Ditto

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,973 @@
...
+
+bool Daemon::ApiImpl::modifyLockPlugins(
+        QList<EncryptionPlugin*> encryptionPlugins,
+        QList<StoragePlugin*> storagePlugins,
+        QList<EncryptedStoragePlugin*> encryptedStoragePlugins,
+        const QByteArray &oldEncryptionKey,
+        const QByteArray &newEncryptionKey)
+{
+    bool allSucceeded = true;
+    for (EncryptionPlugin *eplugin : encryptionPlugins) {
```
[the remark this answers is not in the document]
chriadam: Ditto

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,973 @@
...
+{
+    return plugin->setSecret(collectionName,
+                             hashedSecretName,
+                             secret.identifier().name(),
+                             secret.data(),
+                             secret.filterData(),
+                             key);
+}
+
+std::tuple<Result,
```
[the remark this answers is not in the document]
chriadam: Ok, I will do so.

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,973 @@
...
+    Secret secret;
+    bool locked = false;
+    Result pluginResult = plugin->isLocked(identifier.collectionName(), &locked);
+    if (pluginResult.code() != Result::Succeeded) {
+        return std::make_tuple(pluginResult, secret);
+    }
+
+    // if it's locked, attempt to unlock it
+    if (locked) {
+        pluginResult = plugin->setEncryptionKey(identifier.collectionName(), encryptionKey);
```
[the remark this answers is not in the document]
chriadam: Patches are welcome.

daemon/SecretsImpl/pluginfunctionwrappers_p.h
```diff
@@ -0,0 +1,296 @@
...
+
+namespace Sailfish {
+
+namespace Secrets {
+
+namespace Daemon {
+
+namespace ApiImpl {
+
+std::tuple<bool, // found
```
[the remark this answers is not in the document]
chriadam: Yep, I will do that.

daemon/SecretsImpl/pluginfunctionwrappers_p.h
```diff
@@ -0,0 +1,296 @@
...
+           Sailfish::Secrets::Result> modifyLockSpecificPlugin(
+        const QMap<QString, Sailfish::Secrets::StoragePlugin*> &storagePlugins,
+        const QMap<QString, Sailfish::Secrets::EncryptionPlugin*> &encryptionPlugins,
+        const QMap<QString, Sailfish::Secrets::EncryptedStoragePlugin*> &encryptedStoragePlugins,
+        const QString &lockCodeTarget,
+        std::tuple<QByteArray, QByteArray> newAndOldLockCode);
+
+bool lockPlugins(
+        QList<Sailfish::Secrets::EncryptionPlugin*> encryptionPlugins,
```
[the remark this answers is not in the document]
chriadam: Will fix, should be list refs

daemon/SecretsImpl/pluginfunctionwrappers_p.h
```diff
@@ -0,0 +1,296 @@
...
+        QList<Sailfish::Secrets::StoragePlugin*> storagePlugins,
+        QList<Sailfish::Secrets::EncryptedStoragePlugin*> encryptedStoragePlugins);
+
+bool unlockPlugins(
+        QList<Sailfish::Secrets::EncryptionPlugin*> encryptionPlugins,
```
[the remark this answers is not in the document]
chriadam: The values come directly from QMap<QString, Plugin*>::values() which is documented to be a QList().  I will use a list ref instead of by value.

daemon/SecretsImpl/pluginfunctionwrappers_p.h
chriadam: QtConcurrent::run() has a limit of 5 arguments to a given method, and pointer outparams are a BAD idea for concurrent code.  Hence the tuplisation and outparams done this way.

daemon/SecretsImpl/secrets.cpp
chriadam: GCC below certain versions cannot.  Slava complains when we use ">>" instead of "> >" because he tests with old compilers which don't support the syntax.  There is zero cost to using the old syntax which is supported everywhere, so I will continue to use that one.

daemon/SecretsImpl/secrets.cpp
chriadam: Why is what I've done harder to read?

daemon/SecretsImpl/secrets.cpp
chriadam: Why is the current code hard to read?

daemon/SecretsImpl/secrets.cpp
chriadam: Bleh, if such refactoring is required, let's do it in a separate PR, it's completely unrelated to this one.

chriadam:
Thanks for the reviews!  While the style fixes are important, more important is the functionality, and I am worried e.g. that thread interleaving could cause the bookkeeping database to get into a bad state in some cases (e.g. create collection request interleaved with delete collection request on the same collection).  Does anyone have any functional review comments, e.g. on how to solve that problem?

(I've thought of two possibilities: either move all bookkeeping db operations to the secrets I/O thread as well; or have some per-collection semaphore to prevent another collection-specific operation from being started before the other is finished, etc.  Does anyone have any comments on these possibilities, or alternative solutions?)

daemon/SecretsImpl/secrets.cpp
chriadam: And actually the refactoring should be done properly as I've said multiple times, with per-request-type marshalling structs with named parameters etc.

chriadam:
Have also now added the Crypto API side.

However, there is an issue which I need to look into: the InitialiseCipherSessionRequest implementation causes the following warning to be emitted:

QObject: Cannot create children for a parent that is in a different thread.
(Parent is Sailfish::Crypto::Daemon::Plugins::OpenSslCryptoPlugin(0x1fb4874), parent's thread is QThread(0x1f9ae98), current thread is QThread(0x1fec6f8)

Most likely the threadPoolForPlugin() lookup is failing to produce the correct result.  I need to investigate further.

Finally, I think we do need an initialise() method for each plugin after all, as plugins may have thread-specific initialisation required.  In the case of OpenSSL, I still need to implement thread callbacks as described in https://www.openssl.org/blog/blog/2017/02/21/threads/ to prevent threading issues between the three different openssl-dependent plugins we ship.

daemon/SecretsImpl/secrets.cpp
Venemo: Ok!

daemon/SecretsImpl/pluginfunctionwrappers_p.h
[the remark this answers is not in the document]
Venemo: Ok!

daemon/SecretsImpl/secrets.cpp
Venemo: Ok.
I guess the question of how old compilers doesn't belong here, so let's talk about it somewhere else.

daemon/SecretsImpl/secrets.cpp
Venemo: It's a matter of opinion, I guess. No offence, I just think the initializer list is easier to understand.

izh1979:
One general comment. We need to maintain high code quality especially in security-related parts. Making MRs that needs lot of refactoring from the start increases project's technical debt. And it could result in unmanageable code. And that "later" time for refactoring could never happen and we will live with this code several years. So please pay more attention for refactoring copy-paste code etc.

Yes we need new functionality. But we need superb code quality as well.

izh1979:
Earlier when I worked for Jolla a couple years ago there was very strict code review and I would like to maintain quality on the highest possible level. Because it could save a lot of time in future.

izh1979:
But if it is planned to be refactored soon (e.g. as a part of https://github.com/sailfishos/sailfish-secrets/issues/35) then it could be merged.

daemon/SecretsImpl/secrets.cpp
vityafx: I agree with initializer list too. It is more readable.

daemon/SecretsImpl/secrets.cpp
vityafx: At least because of the width of the code. You clearly see what is happening at a glance, without following the text and the indentation.

Venemo:
@izh1979 I believe all of us agree with you on code quality, and we are doing our best. Keep in mind that the code here is a work-in-progress. To @chriadam 's defense: I think he just wants all the features done first, and then do the necessary refactoring.

daemon/SecretsImpl/secrets.cpp
Venemo: This is, again, of course, a matter of opinion. Perhaps we just come from a different background.

To my eyes, the kind of code where each statement fulfills only one clearly understandable purpose is easier to read than a bunch of things mixed into a single statement. As @vityafx says it's easier to glance what's happening.

Also, when you are debugging the code and wish to know what is passed to `request->connection.send()`, then in the simpler case you can just place a breakpoint and then see it visualized in the debugger. Whereas if it is written as a single statement, it is a bit harder to debug.

Like I said, these are just my 20 cents.

chriadam:
The code quality is high.  Most of the review comments I have seen have been strictly style related (some people preferring initialiser lists, some people preferring explicit instantiation and append; some people preferring a hard 80 column limit, some preferring to use more vertically-compact expressions).

Certainly, good style helps create more maintainable code, I agree with that.  And we should of course strive for best style and maintainability.  Of course.  But also understand that every week there are new features being asked for (e.g. in the last two weeks I have been made aware of plugin lock/unlock requirement, and plugin threading requirement, from OMP side) with no time allowed for refactoring tasks which we know have to be done (and which have been documented! I added the note about the request parameter type marshalling in the README two months ago!!!) and ZERO PRs from OMP to help with that effort.

I understand that your comments are coming from a good place (that is, you care about the quality of the code), and I appreciate that.  But I also care about code quality, and it's somewhat insulting that you assume that I don't, without considering the broader context in which the work occurs.  If you wish to discuss this further, let us do so via email or IRC, perhaps.

chriadam:
Also there is a dimension of code quality which is far more impactful than what has been commented on so far, and that is the correctness and logic of the code.  I have concerns about those aspects with regards to interleaving of bookkeeping database operations due to the threaded nature of the request handling, and yet there have been no comments / suggestions / analysis from anyone about these important functional requirements.  I prefer to focus on functionality in broad strokes, then polish when we agree on the architecture of the feature; from the reviews in the past, it seems like you guys prefer the opposite approach (and that's fine - just we need to accept that there are differences of opinion about how to approach iterative development, and try our best to work with each other to achieve the best outcomes).

daemon/SecretsImpl/secrets.cpp
chriadam: We don't have an 80-column restriction, and IMO for good reason.  Indentation is an important part of code style, indeed there are languages like python which rely solely on it to convey and enforce expression block semantics.  But I agree that some styles of indentation are more readable than others, and you may very well be correct that this style is less readable than the one you suggest.

Regarding splitting over multiple lines, it is somewhat less atomic (e.g. there exists potential for a separate, unrelated instruction with side-effects to be inserted there; alternatively, implies that the variable might be used elsewhere), whereas the syntax I use makes it syntactically clear that the thing being created is solely for the argument and should have no side effects.

Regardless, you may very well be right.  Nonetheless, I would not change this single instance (or the multiple - but still a tiny subset of the total - instances) in this PR because then we would be introducing style inconsistencies with the rest of the code base (which IMO is even worse for maintainability). IMO the correct way to resolve this, if we agree that it should be changed, would be to change such style across the entire project, in a single, atomic commit / PR.  As I have said many times: your PRs are welcome.  From OMP or from Venemo or from community contributors.

chriadam:
I've now added the thread locking callbacks for openssl plugins to ensure thread safety.  I tested using the following script:

```
#!/usr/bin/perl

for $i (1..100)
{   
    if (fork == 0)
    {
        exec ("/opt/tests/Sailfish/Crypto/tst_cryptorequests randomData generateKeyEncryptDecrypt");
        exit;
    }
}
```

(Obviously, the stored-key tests cannot be run in the same way as those other two, or they (expectedly) fail due to the fact that each test tries to create exactly the same collection.)

This all passes without crashing the daemon, which suggests that the thread locking stuff is working appropriately, however of course thorough and careful review would be appreciated.

chriadam:
Updated test which uses separate threads to run things concurrently (and uses sleeps to try to prevent concurrent-runs of the devicelockCollectionSecret tests (because concurrent runs of that one will expectedly fail due to the duplicated collection)).  Everything passes for me on device, with no crashes.  I will run with helgrind also.

```
#!/usr/bin/perl

for $i (1..30)
{
    if (fork == 0)
    {
        exec ("/opt/tests/Sailfish/Crypto/tst_cryptorequests randomData generateKeyEncryptDecrypt");
        exit;
    }
    if (($i % 10) == 0)
    {
        if (fork == 0)
        {
            sleep($i/5);
            exec ("/opt/tests/Sailfish/Secrets/tst_secretsrequests devicelockCollectionSecret");
            exit;
        }
    }
}

exit;
```

izh1979:
@chriadam I didn't want to insult anyone. Sorry if that's happened.
I understand that it's hard to develop complex things when deadlines are near.

chriadam:
Rebased, updated with a very primitive protection against interleaved requests.  Added unit test to ensure that the interleaving protection works correctly.  Comments/suggestions welcome - I'm sure there's a better way to do this but this way at least is simple...

chriadam:
Next I will look at refactoring the plugin API to allow for the improvements noted by vityafx, and rebase these changes on top of it.

chriadam:
Plugin refactoring done, code duplication noted by vityafx should now be resolved for the most part.

The recent GCM support PR has now caused conflicts, I will need to fix those tomorrow.

chriadam:
Rebased, conflicts fixed.  Review appreciated.

daemon/CryptoImpl/cryptopluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,305 @@
...
+{
+    bool verified = false;
+    Result result = plugin->verify(
+                signature, data, key, std::get<0>(options), std::get<1>(options), &verified);
+    return ValidatedResult(result, verified);
+}
+
+TagDataResult CryptoPluginWrapper::encrypt(
+        CryptoPlugin *plugin,
+        std::tuple<QByteArray, QByteArray> dataAndIv,
```
vityafx: `tuple` of two arguments is never better than the `std::pair`. And a struct with named fields is always more readable over `std::pair`.

daemon/CryptoImpl/cryptopluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,305 @@
...
+                &plaintext, &verified);
+    return VerifiedDataResult(result, plaintext, verified);
+}
+
+CipherSessionTokenResult CryptoPluginWrapper::initialiseCipherSession(
+        CryptoPlugin *plugin,
+        quint64 clientId,
+        const QByteArray &iv,
+        const Key &key, // or keyreference, i.e. Key(keyName)
+        std::tuple<CryptoManager::Operation,
```
vityafx: Could we use structs?

daemon/controller.cpp
```diff
@@ -108,6 +108,17 @@ Sailfish::Secrets::Daemon::Controller::crypto() const
     return m_crypto;
 }
 
+QWeakPointer<QThreadPool> Sailfish::Secrets::Daemon::Controller::threadPoolForPlugin(const QString &pluginName) const
+{
+    if (m_secrets->potentialCryptoStoragePlugins().contains(pluginName)) {
```
vityafx: Does the order matter here?

daemon/CryptoImpl/cryptopluginfunctionwrappers.cpp
vityafx: ... And everywhere in the code - there are a lot of tuples.

chriadam: Ok, I will update to use structs.

daemon/controller.cpp
chriadam: Yes.  If a crypto plugin is also an encrypted storage plugin, then all its operations must be run on the Secrets Plugin thread.

plugins/sqlcipherplugin/sqlcipherplugin.h
```diff
@@ -57,8 +59,8 @@ class CryptoPlugin : public Sailfish::Crypto::CryptoPlugin
 };
 
 class Q_DECL_EXPORT SqlCipherPlugin : public QObject
-                                    , public Sailfish::Secrets::Daemon::Plugins::EncryptedStoragePlugin
-                                    , public Sailfish::Secrets::Daemon::Plugins::CryptoPlugin
+                                    , public virtual Sailfish::Secrets::Daemon::Plugins::EncryptedStoragePlugin
```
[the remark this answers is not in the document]
chriadam: Because the SqlCipher plugin derives from both CryptoPlugin and EncryptedStoragePlugin, and this commit refactors the Plugin types to derive from a common base class (as per your and Venemo's review feedback), thus we have diamond multiple inheritance.  The most-base class is ABC but nonetheless virtual inheritance ensures we have only a single base instance.

chriadam:
While the current request interleaving protection seems to work according to the unit testing, I am currently working on a PR to greatly improve the code (by removing the "global" bookkeeping db, and instead having per-plugin bookkeeping dbs which can be updated transactionally with the plugin data, to improve atomicity).  A side-effect of that refactoring is that the (per-plugin) bookkeeping database will now be updated within the plugin thread, and thus no interleaving on the main thread will be possible.

chriadam:
All unit tests pass, I've rebased on top of current master again.

A variety of other improvements are included as part of the refactor; for example, it supports importing keys from plugins at unlock time, and retrieving stored key identifiers from plugins, as requested by IZh and xDen on IRC.

If there are no dissenting voices, we'll merge this tomorrow or Sunday, before the RC branch.

Denis-Semakin:
@chriadam An example of how to retrieve a stored Key from plugin (e.g. token in my case) would be kindly appreciated.

tests/Secrets/tst_secretsrequests/tst_secretsrequests.cpp
```diff
@@ -275,7 +275,11 @@ void tst_secretsrequests::devicelockCollectionSecret()
     QCOMPARE(gsrss.count(), 2);
     QCOMPARE(gsr.status(), Request::Finished);
     QCOMPARE(gsr.result().code(), Result::Succeeded);
-    QCOMPARE(gsr.secret(), testSecret);
+    QCOMPARE(gsr.secret().data(), testSecret.data());
+    QCOMPARE(gsr.secret().type(), testSecret.type());
+    QCOMPARE(gsr.secret().filterData(), testSecret.filterData());
+    QCOMPARE(gsr.secret().name(), testSecret.name());
+    QCOMPARE(gsr.secret().collectionName(), testSecret.collectionName());
```
jpetrell: This is a bit unfortunate. At least create shared function to compare secrets?

jpetrell:
The branch builds without issues and most autotests pass, but unfortunately I get half-dozen autotest case fails and many failures seem to relate to the functionality introduced here (interleaving, plugin threading, initialisation of metadata plugins). On subsequent run more secrets test cases start failing like something broke in the daemon side during the first run.

chriadam:
I had no autotest failures on my L500D device (although the qml test didn't run due to missing some .qml file, I guess a packaging issue?).  Did you completely remove your Secrets folder before attempting the run?

We are out of time and this needs to be in before the RC branch.

tests/Secrets/tst_secretsrequests/tst_secretsrequests.cpp
chriadam: Sure, let's do that - after the branch.  This should not block this PR.

Venemo:
I just started reviewing this. It's a **huge** changeset!

First few things I noticed:
* It doesn't compile on OpenSSL 1.1 (I'm fixing this and will push into this branch.)
* There are several new warnings: one fallthrough (which seems to be a bug in key import), and some others in tests which seem to be typos. I'm fixing these too.

chriadam:
@Denis-Semakin e.g. https://github.com/sailfishos/sailfish-secrets/blob/master/tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp#L937 but yes I will add explicit examples and improved documentation next week.

Venemo:
This seems like a typo in `tst_cryptorequests.cpp`:

warning: logical not is only applied to the left hand side of comparison [-Wlogical-not-parentheses]
warning: comparison of constant ‘2’ with boolean expression is always false [-Wbool-compare]

```
        if (!er.status() == Request::Finished) {
            er.startRequest();
        }
```

I'm fixing it to:

```
        if (er.status() != Request::Finished) {
            er.startRequest();
        }
```

Venemo:
After the above fixes I was able to compile it on my laptop and start the daemon in test mode.

Crypto tests:
* tst_crypto: all pass
* tst_cryptorequests: several failures, daemon segfaults and database is left corrupted
* tst_cryptosecrets: all fail
* tst_evp: all pass
* tst_secrets: several failures

Did not continue to test on my laptop after that. Now compiling with the Sailfish platform SDK to find out how it goes on the device.

chriadam:
I'm checking on my L500D in the office now.  Are you sure you removed the privileged/Secrets folder before starting?  The database layout has completely changed.

Venemo:
@chriadam Yes, I removed that folder. I think it somehow works differently on the laptop, but didn't yet have time to find out why.

On the device, I'm getting several of these warnings with `tst_cryptorequests::storedKeyRequests`. Despite this, the tests pass (though they are very slow on the device). To me it seems like this is a swallowed error.

```
[W] unknown:0 - Cannot deserialise key, bad magic number: 0
[W] unknown:0 - Unable to finish unknown request: 207
```

chriadam:
As discussed on IRC, the unit tests all pass on device.  I fixed the one test which didn't.

Since no-one had any concerns about the architecture or API changes, and since OMP require the RC to contain these changes, I am merging this now.

PR #48, opened 2018-03-26
[sailfish-crypto] Add GCM support to Encrypt/Request & DecryptRequest. Contributes…

blammit:
… to JB#40058


blammit:
The tst_cryptorequests storedKeyRequests() and storedDerivedKeyRequests() tests currently fail, so this patch isn't finished yet. Need to investigate why those fail when generateKeyEncryptDecrypt() passes, but reviews are appreciated in the meantime. I've started adding a similar patch for decryption as well.

plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -240,6 +240,103 @@ int osslevp_aes_decrypt_ciphertext(const EVP_CIPHER *evp_cipher,
...
+            || auth == NULL || auth_length <= 0
+            || key_length <= 0 || key == NULL || encrypted == NULL || tag_length <= 0) {
+        /* Invalid arguments */
+        fprintf(stderr, "%s\n", "invalid arguments, aborting encryption");
+        return -1;
+    }
+
+    /* Allocate the buffer for the encrypted output */
+    ciphertext = (unsigned char *)malloc(ciphertext_length);
+    memset(ciphertext, 0, ciphertext_length);
```
Venemo: Consider using `OPENSSL_zalloc` for this.

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
```diff
@@ -1060,3 +1102,53 @@ Daemon::Plugins::OpenSslCryptoPlugin::aes_decrypt_ciphertext(
...
+                                                       key.size(),
+                                                       (const unsigned char *)auth.constData(),
+                                                       auth.size(),
+                                                       (const unsigned char *)plaintext.constData(),
+                                                       plaintext.size(),
+                                                       &encrypted,
+                                                       &tag,
+                                                       tagSize);
+
+    if (encryptedSize <= 0) {
```
Venemo: Doesn't `-1` from `osslevp_aes_auth_encrypt_plaintext` indicate an error?

plugins/opensslcryptoplugin/evp/evp.c
blammit: Yes sounds good; perhaps we could fix this across the codebase in a separate patch.

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
blammit: Yes, so this returns a null QByteArray in that case.

blammit:
Fixed remaining encryption issues and autotests pass now, but probably best to get decryption finished as well before merging so that the autotests pass for that part of the process as well.

blammit:
Decryption support now implemented as well; ready for review.

plugins/opensslcryptoplugin/evp/evp.c
Venemo: Ok!

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
Venemo: Shouldn't there be an indication of success/failure to the caller? Or does an empty `QByteArray` always mean failure?

blammit: Ideally, all main points of failure would have been checked before calling this aes_auth_encrypt_plaintext() function, so that the only real error that occurs here are to do with a failure to encrypt from the openssl libraries, which is indicated by the empty QByteArray. But, I think having the various checks in this method before calling osslevp_aes_auth_encrypt_plaintext() make this less clear, so I'll change it to move the checks to the calling execEncrypt() method.

Later on, we should probably change the evp.c methods to return appropriate error codes instead of just a -1 on error, to provide more precise error information.

blammit: Patch updated to clean up the argument checking before calling aes_auth_encrypt_ciphertext() and aes_auth_decrypt_ciphertext() and return more precise errors.

lib/Crypto/decryptrequest.cpp
```diff
@@ -252,13 +310,22 @@ void DecryptRequest::startRequest()
...
         }
 
-        QDBusPendingReply<Result, QByteArray> reply =
-                d->m_manager->d_ptr->decrypt(d->m_data,
-                                             d->m_initialisationVector,
-                                             d->m_key,
-                                             d->m_blockMode,
-                                             d->m_padding,
-                                             d->m_cryptoPluginName);
+        QDBusPendingReply<Result, QByteArray> reply = (d->m_blockMode == CryptoManager::BlockModeGcm)
```
chriadam: I wonder about this.  One of the requirements from OMP is that we must support arbitrary plugin-specific block modes (i.e. custom algorithm, custom block mode; see CryptoManager::BlockModeCustom).  It could be that such a block mode is authenticated even though it isn't one of GCM or CCM, and thus we must not switch the service call based on this.

That is: instead of adding a separate "authenticatedDecrypt" (and authenticatedEncrypt) which we call if and only if the block mode is a "known" authenticated mode, we should probably simply expand the "normal" encrypt/decrypt APIs so that they take/return the authenticated operation data, and simply document that those extra authenticated operation data are not valid for non-authenticated block modes (such as GCM/CCM).

blammit: What if this if condition is changed to check for whether authenticationData is set, rather than checking the block mode? Is that sufficient to support the arbitrary block mode requirement?

Whether we should have both encrypt() and authenticatedEncrypt() methods in the daemon/plugins can be considered separately; I think having a separate authenticatedEncrypt() is useful to be explicit about the fact that the authentication data and tag are required for the auth case.

blammit:
Updated EncryptRequest and DecryptRequest to do authentication based on whether authenticationData is set, instead of checking the blockMode value.

daemon/CryptoImpl/crypto.cpp
```diff
@@ -1234,14 +1367,39 @@ void Daemon::ApiImpl::CryptoRequestQueue::handleFinishedRequest(
...
             break;
         }
+        case AuthenticatedEncryptRequest: {
+            Result result = request->outParams.size()
+                    ? request->outParams.takeFirst().value<Result>()
+                    : Result(Result::UnknownError,
+                             QLatin1String("Unable to determine result of AuthenticatedEncryptRequest request"));
+            if (result.code() == Result::Pending) {
+                // shouldn't happen!
+                qCWarning(lcSailfishCryptoDaemon) << "EncryptRequest:" << request->requestId << "finished as pending!";
```
chriadam: this should be AuthenticatedEncryptRequest not EncryptRequest

daemon/CryptoImpl/crypto.cpp
```diff
@@ -1234,14 +1367,39 @@ void Daemon::ApiImpl::CryptoRequestQueue::handleFinishedRequest(
...
+                        : QByteArray();
+                request->connection.send(request->message.createReply() << QVariant::fromValue<Result>(result)
+                                                                        << QVariant::fromValue<QByteArray>(encrypted)
+                                                                        << QVariant::fromValue<QByteArray>(tag));
+                *completed = true;
+            }
+            break;
+        }
+        case AuthenticatedDecryptRequest:
+            // fall through, result/parameters are the same as for non-authenticated
```
chriadam: Are they?  Shouldn't it also produce a boolean "verified" out-parameter?

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -299,6 +304,13 @@ void tst_cryptorequests::generateKeyEncryptDecrypt()
     // test encrypting some plaintext with the generated key
     QByteArray plaintext = "Test plaintext data";
     QByteArray initVector = "0123456789abcdef";
+    QByteArray authData("fedcba9876543210");
+    QByteArray tag;
+
+    if (blockMode == CryptoManager::BlockModeGcm) {
+        initVector.resize(12);  // GCM default IV size is 12
```
chriadam: Why is this required for these tests, but not the cipherEncryptDecrypt() one?

blammit: I think that is a bug in the cipherEncryptDecrypt() case. It should check the IV size as well. Will look into that.

daemon/CryptoImpl/crypto.cpp
blammit: According to https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption checking the result of EVP_DecryptFinal_ex() is sufficient, or have I missed something?

blammit: @chriadam Could you clarify the purpose of the "verified" return value for finaliseCipherSession() in the plugins? I would have thought that returning an "error" result would be sufficient if the EVP_DecryptFinal_ex() operation fails, instead of requiring an additional "verified" value (assuming that the decrypted data should just be discarded if this verification fails).

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
blammit: Ah actually, it's because OpenSslCryptoPlugin::initialiseCipherSession() automatically generates a new IV if the specified IV has an incorrect size. I think we discussed that Encrypt/DecryptRequest would just use the given IV instead of trying to correct it, but we should probably have a consistent approach across both methods?

plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
Venemo: Cool, thank you. Agreed about updating evp.c later.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
Venemo: When it generates a new IV, how does the user get notified about the new IV?

daemon/CryptoImpl/crypto.cpp
```diff
@@ -323,6 +323,36 @@ void Daemon::ApiImpl::CryptoDBusObject::decrypt(
...
+        const QByteArray &authenticationData,
+        const QByteArray &tag,
+        const QString &cryptosystemProviderName,
+        const QDBusMessage &message,
+        Result &result,
+        QByteArray &decrypted)
+{
+    Q_UNUSED(decrypted);  // outparam, set in handlePendingRequest / handleFinishedRequest
+    QList<QVariant> inParams;
+    inParams << QVariant::fromValue<QByteArray>(data);
```
Venemo: Not a problem, just a question: is the template argument necessary here? Can't the compiler figure it out on its own without explicity specifying it?

plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -336,6 +336,108 @@ int osslevp_aes_auth_encrypt_plaintext(const EVP_CIPHER *evp_cipher,
...
+        /* Invalid arguments */
+        fprintf(stderr,
+                "%s: %s\n",
+                "osslevp_aes_auth_decrypt_ciphertext()",
+                "invalid arguments, aborting decryption");
+        return -1;
+    }
+
+    /* Allocate the buffer for the decrypted output */
+    plaintext = (unsigned char *)malloc(ciphertext_length + AES_BLOCK_SIZE);
```
Venemo: Shouldn't the allocation happen after you know the actual decrypted size from `EVP_DecryptFinal_ex`?

plugins/opensslcryptoplugin/evp/evp.c
```diff
@@ -336,6 +336,108 @@ int osslevp_aes_auth_encrypt_plaintext(const EVP_CIPHER *evp_cipher,
...
+                                        const unsigned char *init_vector,
+                                        const unsigned char *key,
+                                        int key_length,
+                                        const unsigned char *auth,
+                                        int auth_length,
+                                        unsigned char *tag,
+                                        int tag_length,
+                                        const unsigned char *ciphertext,
+                                        int ciphertext_length,
+                                        unsigned char **decrypted)
```
Venemo: Shouldn't there be an out parameter `decrypted_length`?

Venemo: Ah, I just now noticed that it's the return value. Sorry.

plugins/opensslcryptoplugin/evp/evp.c
blammit: Hm, then how would you pass an allocated buffer to EVP_DecryptUpdate() to get the decrypted result?

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
blammit: For CipherRequest, the IV can be checked via CipherRequest::generatedInitialisationVector.

daemon/CryptoImpl/crypto.cpp
blammit: It's not necessary for Qt types like QByteArray and QString; it used to be required for custom registered types, but I'm not sure of the requirements for current Qt/compiler configurations.

daemon/CryptoImpl/crypto.cpp
chriadam: In GCM mode, the EVP_DecryptFinal_ex() value does not produce additional plaintext, instead it produces (in effect) a boolean "did the calculated tag match the provided tag" result.  The decrypted data (plaintext) should still be returned to the client (although in practice the client should not use that returned plaintext, since it is guaranteed to have been generated from corrupted ciphertext).

chriadam: Also, note that no error has occurred - GCM has worked successfully in this case - that is, it successfully completed and determined that the plaintext should not be trusted.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
chriadam: I agree that we should be consistent across both methods and offer generatedInitialisationVector() for both.  However, perhaps the semantics of that should be: if any IV is provided, but is wrong, return an error; only generate an IV if an empty byte array is provided as the input IV.  Does this sound sensible or are there good reasons not to do this?

daemon/CryptoImpl/crypto_p.h
```diff
@@ -391,6 +391,32 @@ public Q_SLOTS:
             Sailfish::Crypto::Result &result,
             QByteArray &decrypted);
 
+    void authenticatedEncrypt(
```
chriadam: After more thought, I'm not convinced that this should be a separate method.  It's still encryption and decryption, using the same algorithms as anything else, only the mode is different (and some different parameters required).  We can document those parameter differences in the request object, so not sure having separate methods makes sense.

daemon/CryptoImpl/crypto.cpp
blammit: Okay, I see. Would it make sense to return some other error, e.g. AuthenticationVerificationFailed or something instead of a 'Success' code, and then still return the decrypted plaintext?

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
blammit: What if we have a generateInitializationVector() or such, so that the user can explicitly assign one? That way there is no uncertainty about whether one has been automatically generated.

(Btw I noticed we sometimes use "initialise" and other times "initialize", so that should be standardized soonish, presumably to the US "initialize" spelling.)

daemon/CryptoImpl/crypto.cpp
chriadam: Do you think the Verify operation should produce a boolean "verified" outparam, or just return a VerificationFailed vs Success code in the result?  Whatever we decide, we should be consistent across all operations.

Personally, I'm partial to using explicit outparams, and using error codes only for actual errors (that is, something went wrong), but I'm willing to be convinced that that is less intuitive for users or otherwise suboptimal.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
chriadam: Yes, we should standardise to American english.

I don't think we should have a separate generateInitializationVector operation/request - it seems far more intuitive/simple to me to have it generated as part of the encrypt request, honestly (mainly because it's not a separate thing - the IV is part of the encrypt operation.  We provide a generateRandomData request for the case where they want to explicitly generate an IV separately and set it).  What are your reasons for preferring a separate request for that?

daemon/CryptoImpl/crypto_p.h
blammit: Originally I tried just extending the encrypt() method and the code became convoluted in checking the different paths and conditions, so it seemed better to just separate the functionality. The explicit separation in the API seems more clear to the user as well, instead of using parameter checking to determine the correct operation. 

For the D-Bus method, the authenticatedEncrypt() returns an additional parameter (the tag) so that one can't be merged with encrypt(), unless encrypt() is also changed to return additional (empty) values.

blammit:
Added various changes as discussed with Chris, and rebased against latest master.

blammit:
Yes good point. Renamed to 'authenticationTag' and rebased against master.

chriadam:
LGTM!

PR #51, opened 2018-04-01
Refactor / fix plugin load mechanism

Venemo:


daemon/plugin_p.cpp
```diff
@@ -6,43 +6,99 @@
...
-void Daemon::ApiImpl::PluginHelper::reportFailure(
-        const QLoggingCategory &category())
+void Daemon::ApiImpl::PluginManager::addPlugin(QPluginLoader *loader, const PluginHelpers::PluginInfo &info, QObject *obj)
 {
-    if (m_failureType != PluginLoadFailure
-            && m_failureType != PluginTypeFailure) {
-        // already reported this error
-        return;
+    bool use = true;
+    bool testMode = isAutotestMode();
```
chriadam: can be declared as const I think

daemon/plugin_p.cpp
```diff
@@ -6,43 +6,99 @@
...
+
+    if (!info.canUse) {
+        qWarning() << "Not a useable crypto or secrets plugin:" << loader->fileName();
```
chriadam: typo, should be "usable"

daemon/plugin_p.cpp
```diff
@@ -6,43 +6,99 @@
...
+        use = false;
+    }
+
+    if (m_plugins.contains(info.name)) {
+        qWarning() << "Not adding plugin with duplicate name:" << loader->fileName();
```
chriadam: I think we should use qCWarning() with an appropriate logging category instead of raw qWarning

daemon/CryptoImpl/cryptorequestprocessor.cpp
```diff
@@ -88,63 +90,6 @@ Daemon::ApiImpl::RequestProcessor::RequestProcessor(
             this, &Daemon::ApiImpl::RequestProcessor::secretsCryptoPluginLockCodeRequestCompleted);
 }
 
-bool
-Daemon::ApiImpl::RequestProcessor::loadPlugins()
-{
-    // First, see if any of the EncryptedStorage plugins from Secrets are also
-    // Crypto plugins (providing generateAndStoreKey() functionality internally).
-    qCDebug(lcSailfishCryptoDaemon) << "Loading crypto storage plugins";
-    QMap<QString, QObject*> potentialCryptoPlugins = m_secrets->potentialCryptoStoragePlugins();
```
chriadam: Does the new code handle these correctly?  A quick grep didn't find any calls to potentialCryptoStoragePlugins() in the new code, but I might have missed something.

chriadam:
We should also refactor the Secrets/extensionplugins.h declarations so that there is a base SecretsPlugin type, with the various locking + metadata virtual methods.  That would allow refactor/cleanup of the various locking/unlocking/etc code, as pointed out by vityafx in the thread serialisation review.

daemon/plugin_p.cpp
Venemo: Ok!

daemon/plugin_p.cpp
Venemo: Will fix.

daemon/plugin_p.cpp
Venemo: Which category should it be?

daemon/CryptoImpl/cryptorequestprocessor.cpp
Venemo: These plugins are loaded and should be returned by both `getPlugins<CryptoPlugin>()` and `getPlugins<EncryptedStoragePlugin>()` since they can be `qobject_cast()`ed to both. Is there any more special handling necessary?

daemon/plugin_p.cpp
vityafx: I'd suggest something like "sailfishsecrets.plugin-loader" ?

daemon/CryptoImpl/cryptorequestprocessor.cpp
chriadam: Yes.  See the threading PR for a good example: we need to know which Crypto plugins are also EncryptedStoragePlugins so that we can run operations interacting with those plugins on the appropriate (in that case, Secrets) thread.  There may be other cases in the code which currently call the potentialCryptoStoragePlugins() method also.

daemon/plugin_p.cpp
Venemo: Ok.

Venemo: Currently we have `org.sailfishos.crypto.daemon` and `org.sailfishos.secrets.daemon`.

Venemo:
FIxed the issues raised in review, added support for potential crypto-storage plugins. @chriadam please take a look if it's OK now.

daemon/plugin_p.cpp
```diff
@@ -6,43 +6,105 @@
...
+        Q_FOREACH (const QFileInfo &file, dir.entryInfoList(QDir::Files | QDir::NoDot | QDir::NoDotDot, QDir::Name)) {
+            const QString fileName = file.fileName();
+
+            // Don't even try to load files which don't look like libraries
+            if (!fileName.startsWith("lib") || !fileName.contains(".so")) {
+                continue;
+            }
+
+            // load the plugin
+            auto *loader = new QPluginLoader(file.absoluteFilePath());
```
chriadam: Is this leaked in the case where the plugin is used?

daemon/plugin_p.cpp
```diff
@@ -6,43 +6,105 @@
...
-    if (!isLoaded()) {
-        qCWarning(category) << "cannot load plugin:"
-                            << fileName() << errorString();
+    if (!use) {
+        if (!loader->unload()) {
+            qCWarning(lcSailfishSecretsPlugins) << "Could not unload plugin:" << loader->fileName();
+        }
+        delete loader;
         return;
     }
```
chriadam: Maybe add an else block with: { qCDebug() << "Adding plugin:" << loader->fileName(); delete loader; }

chriadam: Or perhaps "Adding plugin:" << loader->fileName() << "as" << info.name; for maximum information.

chriadam:
LGTM thanks!

PR #56, opened 2018-04-05
[sailfishsecrets] Start sailfish-secretsd upon boot. Fixes JB#41547

rainemak:
This should be enough. However, had some difficulties to verify that all is good. Having "--test" in service file was not enough to verify this.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -9,6 +9,10 @@
 #include <QtCore/QLoggingCategory>
 #include <QtCore/QDir>
 
+#include <QTimer>
+#include <systemd/sd-daemon.h>
+#include <getdef.h>
+
 #include "controller_p.h"
 #include "logging_p.h"
 #include "plugin_p.h"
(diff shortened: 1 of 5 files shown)


daemon/sailfish-secretsd.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Sailfish Secrets Daemon
+Requires=dbus.socket
+After=local-fs.target dbus.socket init-done.service
+Conflicts=shutdown.target
```
Denis-Semakin: May be also add actdead.target in "Conflicts"?

daemon/sailfish-secretsd.service
```diff
@@ -0,0 +1,17 @@
...
+
+[Service]
+Type=notify
```
saukko: Why type notify instead of dbus? This service registers entry to dbus, so why don't we use that? Also we should have dbus unit file to bind the dbus service for this systemd service.

The thing here is that if we do not bind the systemd service and dbus unit, things can start this service through dbus eventhough the status would be conflicting e.g. we are heading towards shutdown.target. Which has lead to issues in the past. With binding if someone would try to start this service over dbus it would be prevented by systemd as it maps the same rules than in normal systemd start to dbus start.

Example of how binding can be done.
https://github.com/nemomobile/ssu/pull/35/files

daemon/main.cpp
```diff
@@ -23,6 +27,16 @@ Q_LOGGING_CATEGORY(lcSailfishCryptoDaemonDBus, "org.sailfishos.crypto.daemon.dbu
 
 Q_DECL_EXPORT int main(int argc, char *argv[])
 {
+    // HACK: not part of actual session so hardcode the session bus socket
```
jpetrell: Refer to Bug id?

PR #66, opened 2018-04-12
[sailfish-secrets] Fix libcrypto linking to sqlcipherplugin and opensslplugin. Contributes to JB#41547

rainemak:
Now sailfish-secrets can be started without test packages being installed as service.


jpetrell:
LGTM

PR #68, opened 2018-04-12
Replace master bookkeeping database with per-plugin metadata dbs

chriadam:
OMP recently gave requirement that we must support importing keys from
plugins.  Plugins can ship with different collections and keys already
defined.

Given that some plugins (e.g. backed by USB tokens) might define
identical and specific collection names, we cannot adequately ensure
that secretName+collectionName pairs are unique without "breaking"
the behaviour the plugin would document for its clients.

Furthermore, some plugins might source data from elsewhere (e.g.
in the case of gpgme, database entries might come from other processes,
or in the case of some online-backed store, from the online service).

Finally, some plugins might not (or might not be able to) export any
information about available collections/keys on the device until the
device has been unlocked via an asynchronous user prompt, which
provides opportunity for a collection with a conflicting name to be
added into a separate plugin prior to unlock occuring.

To resolve these issues, we now require a secret (or key) to be
identified by a tuple of secretName+collectionName+storagePluginName.
Metadata for each plugin is stored in per-plugin databases, which
allows transactions to be used to ensure metadata and plugin-stored
data are modified atomically, and all on the same thread, to avoid
multithreading issues (e.g. metadata modification interleaves) and to
mitigate abrupt power loss failures.


chriadam:
Compiles, most of the secrets unit tests pass, a couple still fail which I need to fix.  I also need to rebase on top of PR47 etc (and fix that one to also cover the importStoredKey case with threading).  But mostly there.

chriadam:
All unit tests now pass.  I need to rebase threadplugins branch on top of current master, and implement the importKey+importStoredKey via QtConcurrent invocation, then I can rebase apifixes on top of that, and this PR on top of apifixes, and then we're good to go.

daemon/SecretsImpl/secretsrequestprocessor_p.h
```diff
@@ -249,6 +250,11 @@ class RequestProcessor : public QObject
             quint64 requestId,
             const Sailfish::Secrets::InteractionParameters &uiParams);
 
+    void preventInterleavedRequests(const QString &collectionName);
+    void allowInterleavedRequests(const QString &collectionName);
+    bool interleavedRequestsAllowed(const QString &collectionName) const;
```
jpetrell: isInterleavedRequestAllowed

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -2263,5 +2264,80 @@ void tst_cryptorequests::lockCode()
                                                     .arg(DEFAULT_TEST_CRYPTO_PLUGIN_NAME));
 }
 
+
+void tst_cryptorequests::pluginThreading()
+{
+    // This test is meant to be run manually and
+    // concurrently with tst_secretsrequests::pluginThreading().
```
jpetrell: How? Manual setup required?

daemon/CryptoImpl/cryptopluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,271 @@
...
+#include "cryptopluginfunctionwrappers_p.h"
+#include "logging_p.h"
+
+using namespace Sailfish::Crypto;
+using namespace Sailfish::Crypto::Daemon::ApiImpl;
+
+/* These methods are to be called via QtConcurrent */
+
+bool CryptoPluginWrapper::isLocked(
+        CryptoPlugin *plugin)
```
jpetrell: The parameter doesn't need to be on second line. Would expect const in such functions, but I guess that could be an issue for some plugins.

daemon/CryptoImpl/cryptopluginfunctionwrappers.cpp
```diff
@@ -0,0 +1,271 @@
...
+    quint32 cipherSessionToken = 0;
+    Result result = plugin->initialiseCipherSession(
+                clientId,
+                iv,
+                key,
+                std::get<0>(options),
+                std::get<1>(options),
+                std::get<2>(options),
+                std::get<3>(options),
+                std::get<4>(options),
```
jpetrell: Scary..

chriadam:
I have now pushed this into PR47.

PR #70, opened 2018-04-16
Implement a gpgme-based crypto plugin allowing to use PGP keys

dcaliste:
Hello @Venemo, @chriadam and @jpetrell. This is a pull request to show the current state of affair on the GnuPG plugin. It's not ready yet and I'm still working on it. It currently contains basics to generate a key and sign with it (not fully working yet though). Later on I'll add the code to create a pinentry as discussed in #5.

It can serve as base for the discussion of tomorrow's meeting though if you are still fit after the sprint of recent days ;)

Congrats for the work by the way.

### DIFF
--- a/plugins/gnupgplugin/gpgme_p.h
+++ b/plugins/gnupgplugin/gpgme_p.h
@@ -1,0 +1,301 @@
+/*
+ * Copyright (C) 2018 Damien Caliste.
+ * Contact: Damien Caliste <dcaliste@free.fr>
+ * All rights reserved.
+ * BSD 3-Clause License, see LICENSE.
+ */
+
+#ifndef GPGME_P_H
+#define GPGME_P_H
+
+#ifndef _FILE_OFFSET_BITS
+#define _FILE_OFFSET_BITS 64
+#endif
+#include <gpgme.h>
+
+#include <Crypto/Plugins/extensionplugins.h>
+#include <Crypto/key.h>
+#include <QString>
+#include <QDebug>
+#include <QDateTime>
+
+struct GPGmeContext {
+    gpgme_ctx_t ctx;
+    gpgme_error_t err;
+    GPGmeContext(gpgme_protocol_t protocol, const QString &home = QString())
+        : ctx(0), err(0)
+    {
+        err = gpgme_engine_check_version(protocol);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            gpgme_engine_info_t info;
+            gpgme_error_t err2;
+            err2 = gpgme_get_engine_info(&info);
+            if (!err2) {
+                while (info) {
+                    if (info->protocol == protocol) {
+                        qCWarning(lcSailfishCryptoPlugin) << "protocol:" << gpgme_get_protocol_name(info->protocol);
+                        if (info->file_name && !info->version)
+                            qCWarning(lcSailfishCryptoPlugin) << "engine" << info->file_name
+                                       << "not installed properly";
+                        else if (info->file_name && info->version && info->req_version)
+                            qCWarning(lcSailfishCryptoPlugin) << "engine" << info->file_name
+                                       << "version" << info->version
+                                       << "installed, but at least version"
+                                       << info->req_version << "required";
+                        else
+                            qCWarning(lcSailfishCryptoPlugin) << "unknow issue";
+                    }
+                    info = info->next;
+                }
+            } else {
+                qCWarning(lcSailfishCryptoPlugin) << "cannot get engine info:" << gpgme_strerror(err2);
+            }
+            return;
+        }
+
+        err = gpgme_new(&ctx);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            ctx = 0;
+            return;
+        }
+        err = gpgme_set_protocol(ctx, protocol);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            gpgme_release(ctx);
+            ctx = 0;
+            return;
+        }
+        if (home.isEmpty()) {
+            return;
+        }
+        gpgme_engine_info_t info = gpgme_ctx_get_engine_info(ctx);
+        while (info && info->protocol != protocol)
+            info = info->next;
+        if (!info) {
+            qCWarning(lcSailfishCryptoPlugin) << "cannot change home.";
+            return;
+        }
+        err = gpgme_ctx_set_engine_info(ctx, protocol, info->file_name,
+                                        home.toUtf8().constData());
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            gpgme_release(ctx);
+            ctx = 0;
+            return;
+        }
+    }
+    ~GPGmeContext()
+    {
+        if (ctx) {
+            gpgme_release(ctx);
+        }
+    }
+    bool operator!()
+    {
+        return (ctx == 0);
+    }
+    operator gpgme_ctx_t() const
+    {
+        return ctx;
+    }
+    QString error() const
+    {
+        return (err && gpgme_err_code(err) != GPG_ERR_NO_ERROR)
+            ? QString(gpgme_strerror(err)) : QString();
+    }
+};
+
+struct GPGmeData {
+    gpgme_data_t data;
+    gpgme_error_t err;
+    GPGmeData()
+        : data(0), err(0)
+    {
+        err = gpgme_data_new(&data);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            data = 0;
+        }
+    }
+    // The data hold by origin are not copied. origin should be valid
+    // for the whole life of the created structure.
+    GPGmeData(const QByteArray &origin)
+        : data(0), err(0)
+    {
+#define NO_COPY 0
+        err = gpgme_data_new_from_mem(&data, origin.constData(),
+                                      origin.length(), NO_COPY);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            data = 0;
+        }
+    }
+    ~GPGmeData()
+    {
+        if (data) {
+            gpgme_data_release(data);
+        }
+    }
+    operator gpgme_data_t() const
+    {
+        return data;
+    }
+    bool operator!()
+    {
+        return (data == 0);
+    }
+    QString error() const
+    {
+        return (err && gpgme_err_code(err) != GPG_ERR_NO_ERROR)
+            ? QString(gpgme_strerror(err)) : QString();
+    }
+    void releaseData(QByteArray *output)
+    {
+        size_t ln;
+        char *signData = gpgme_data_release_and_get_mem(data, &ln);
+        data = 0;
+
+        *output = QByteArray(signData, ln);
+        gpgme_free(signData);
+    }
+};
+
+struct GPGmeKey {
+    gpgme_key_t key;
+    gpgme_subkey_t sub;
+    gpgme_error_t err;
+    enum Level {
+        Public,
+        Secret
+    };
+    GPGmeKey(const gpgme_ctx_t ctx, const char *fingerprint, Level level = Public)
+        : key(0), sub(0), err(0)
+    {
+        err = gpgme_get_key(ctx, fingerprint, &key, level == Secret ? 1 : 0);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            key = 0;
+            return;
+        }
+        if (!key->subkeys) {
+            gpgme_key_unref(key);
+            key = 0;
+            return;
+        }
+        sub = key->subkeys;
+        while (sub && strcmp(fingerprint, sub->fpr))
+            sub = sub->next;
+    }
+    GPGmeKey(const gpgme_ctx_t ctx, const QString &fingerprint, Level level = Public)
+        : GPGmeKey(ctx, fingerprint.toLocal8Bit().constData(), level)
+    {
+    }
+    GPGmeKey(const gpgme_ctx_t ctx)
+        : key(0), sub(0), err(0)
+    {
+        err = gpgme_op_keylist_next(ctx, &key);
+        if (gpg_err_code(err) == GPG_ERR_NO_ERROR) {
+            sub = key->subkeys;
+        } else {
+            key = 0;
+        }
+    }
+    ~GPGmeKey()
+    {
+        if (key) {
+            gpgme_key_unref(key);
+        }
+    }
+    operator gpgme_key_t() const
+    {
+        return key;
+    }
+    bool operator!()
+    {
+        return (key == 0);
+    }
+    operator bool()
+    {
+        return (key != 0);
+    }
+    QString error() const
+    {
+        return (err && gpgme_err_code(err) != GPG_ERR_NO_ERROR)
+            ? QString(gpgme_strerror(err)) : QString();
+    }
+    bool contains(const char *fingerprint) const
+    {
+        if (!key || !fingerprint) {
+            return false;
+        }
+
+        gpgme_subkey_t subkey = key->subkeys;
+        while (subkey) {
+            if (!strcmp(subkey->fpr, fingerprint)) {
+                return true;
+            }
+            subkey = subkey->next;
+        }
+        return false;
+    }
+    const char* fingerprint() const
+    {
+        return sub ? sub->fpr : (const char*)0;
+    }
+    void toKey(Sailfish::Crypto::Key *output, const QString &pluginName) const
+    {
+        if (!key || !key->subkeys || !sub) {
+            return;
+        }
+
+        output->setName(sub->fpr);
+        output->setCollectionName(key->uids->uid);
+        output->setStoragePluginName(pluginName);
+        output->setPublicKey(key->subkeys->fpr);
+
+        QStringList emails;
+        gpgme_user_id_t uid = key->uids;
+        while (uid) {
+            if (uid->email) {
+                emails << uid->email;
+            }
+            uid = uid->next;
+        }
+        output->setFilterData("User-Emails", emails.join(","));
+        output->setFilterData("Expired", sub->expired ? "true" : "false");
+        if (sub->expires) {
+            output->setFilterData("Expire-Date",
+                                  QDateTime::fromMSecsSinceEpoch(sub->expires).toString());
+        }
+        if (sub->timestamp > 0) {
+            output->setFilterData("Creation-Date",
+                                  QDateTime::fromMSecsSinceEpoch(sub->timestamp).toString());
+        }
+
+        output->setOrigin(Sailfish::Crypto::Key::OriginUnknown);
+        output->setSize(key->subkeys->length);
+        switch (key->subkeys->pubkey_algo) {
+        case (GPGME_PK_RSA_E):
+        case (GPGME_PK_RSA_S):
+        case (GPGME_PK_RSA):
+            output->setAlgorithm(Sailfish::Crypto::CryptoManager::AlgorithmRsa);
+            break;
+        case (GPGME_PK_ELG_E):
+        case (GPGME_PK_ELG):
+            output->setAlgorithm(Sailfish::Crypto::CryptoManager::AlgorithmUnknown);
+            break;
+        case (GPGME_PK_DSA):
+            output->setAlgorithm(Sailfish::Crypto::CryptoManager::AlgorithmDsa);
+            break;
+        };
+        Sailfish::Crypto::CryptoManager::Operations op
+            = Sailfish::Crypto::CryptoManager::OperationVerify
+            | Sailfish::Crypto::CryptoManager::OperationDecrypt;
+        if (key->subkeys->can_encrypt) {
+            op |= Sailfish::Crypto::CryptoManager::OperationEncrypt;
+        }
+        if (key->subkeys->can_sign) {
+            op |= Sailfish::Crypto::CryptoManager::OperationSign;
+        }
+        output->setOperations(op);
+
+        output->setPrivateKey("GnuPG");
+    }
+};
+
+#endif
--- a/plugins/gnupgplugin/gpgme_p.h
+++ b/plugins/gnupgplugin/gpgme_p.h
@@ -188,17 +188,44 @@ struct GPGmeKey {
     GPGmeKey(const gpgme_ctx_t ctx)
         : key(0), sub(0), err(0)
     {
-        err = gpgme_op_keylist_next(ctx, &key);
-        if (gpg_err_code(err) == GPG_ERR_NO_ERROR) {
-            sub = key->subkeys;
-        } else {
-            key = 0;
-        }
+        next(ctx);
+    }
+    GPGmeKey(gpgme_error_t err)
+        : key(0), sub(0), err(err)
+    {
+    }
+    GPGmeKey(const GPGmeKey &other)
+        : key(other.key), sub(other.sub), err(other.err)
+    {
+        if (key)
+            gpgme_key_ref(key);
     }
     ~GPGmeKey()
     {
         if (key) {
             gpgme_key_unref(key);
+        }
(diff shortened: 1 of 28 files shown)


tests/plugins/testgnupgplugin/tst_gnupgplugin.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+void tst_gnupgplugin::storedKeyIdentifiers()
+{
+    StoredKeyIdentifiersRequest req;
+
+    req.setManager(&cm);
+    QSignalSpy reqstat(&req, &StoredKeyIdentifiersRequest::statusChanged);
+    QSignalSpy reqiden(&req, &StoredKeyIdentifiersRequest::identifiersChanged);
+
+    // req.setCryptoPluginName(OPENPGP_PLUGIN);
```
chriadam: With the refactor, you need to do: setStoragePluginName(OPENPGP_PLUGIN); and also that plugin must also implement the EncryptedStoragePlugin interface (as well as the CryptoPlugin interface) -- see e.g. the SqlCipher plugin for example.

dcaliste: Thanks, I've added implementation for the EncryptedStorage API.

dcaliste:
@chriadam I've updated my PR. It's now creating PGP keys, listing, signing and verifying.

I've some remarks on the API that we can discuss tomorrow.

dcaliste:
For reference, this is how the pinentry dialog looks like when doing a PGP signing operation from command line with ```gpg2``` executable:

Venemo:
Congrats @dcaliste that does look very cool. Thank you for doing this!

dcaliste:
And now, also from the command-line tool:
<pre>
[nemo@Sailfish devel]$ secrets-tool --sign org.sailfishos.crypto.plugin.gnupg.openpgp org.sailfishos.crypto.plugin.gnupg.openpgp "Damien Caliste <dcaliste@free.fr>" F58016BD85212EFF4763BC70E7F32BC76D8C1C73 SHA256 toto
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2.0.4 (GNU/Linux)

iQGVAwUBWwaoqefzK8dtjBxzAQJAvQwAsj5Gynbp4nLbYuKuN2wCQNGKRuxGjP8V
SnqWKzJ9UI9qbZrKmX86wx5mbdT7vjODHqjpu0Ua05/RVn95Gy9F8GrBjDuuEX+N
OflUnf8juwFcgig3m3lcbgyQFcqt9eGFbm41lKZuLh+MD5Q+opRcLA0wG6OLYpr9
30xuOvd74x8wlvrymxP6gslZoOnCelT2lZsCuMdTE01tPvNsVKqq0gduKIBYYWv5
giwO5GMge29f+lwr9BgwZGqcVHrL5Lc7VHITPHtGKqIYzZpA93gDn/sNQ07/F98V
XLuR3ze2z7F3InbChemRdIaQm7bWUiSmh5MdIW0x08ESezWqdlv1FxlcHOOc3dXh
0l2Us9kyTHEH5S4qIhXaY+UBvF544uQgUFYNPOW0JFsIaK72HrKNeNE0fNVuqYHX
dLDEGH5CxcsP2KZvoFtx26kLkbgLZJOsYvE5aIynkCTCaNO1fKcK3W1J1nBLQ5qd
Skz+aXwU1wRPNc4zw2TYMwUwVV/L+2Fr
=TqYy
-----END PGP SIGNATURE-----
</pre>

One can also list the available keys:
<pre>
[nemo@Sailfish devel]$ secrets-tool --list-keys org.sailfishos.crypto.plugin.gnupg.openpgp
--------------------------------------------
Storage plugin: "org.sailfishos.crypto.plugin.gnupg.openpgp"
    Collection: "Mattia Rizzolo <mattia@mapreri.org>"
          Name: "66AE2B4AFCCF3F52DA184D184B043FCDB9444540"
--------------------------------------------
Storage plugin: "org.sailfishos.crypto.plugin.gnupg.openpgp"
    Collection: "Mattia Rizzolo <mattia@mapreri.org>"
          Name: "6E9BB401D6EC87D3FD1F4EAD39446A919948BA4E"
--------------------------------------------
Storage plugin: "org.sailfishos.crypto.plugin.gnupg.openpgp"
    Collection: "Mattia Rizzolo <mattia@mapreri.org>"
          Name: "8B7868786C33E5C64C4D0A480816B9E18C762BAD"
--------------------------------------------
Storage plugin: "org.sailfishos.crypto.plugin.gnupg.openpgp"
    Collection: "Mattia Rizzolo <mattia@mapreri.org>"
          Name: "4A51F45E4DC34E312FC79C9ABC3F58C272FC5DA4"
--------------------------------------------
Storage plugin: "org.sailfishos.crypto.plugin.gnupg.openpgp"
    Collection: "Damien Caliste <dcaliste@free.fr>"
          Name: "F58016BD85212EFF4763BC70E7F32BC76D8C1C73"
</pre>

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,273 @@
...
+            gpgme_engine_info_t info;
+            gpgme_error_t err2;
+            err2 = gpgme_get_engine_info(&info);
+            if (!err2) {
+                while (info) {
+                    if (info->protocol == protocol) {
+                        qWarning() << "protocol:" << gpgme_get_protocol_name(info->protocol);
+                        if (info->file_name && !info->version)
+                            qWarning() << "engine " << info->file_name
+                                       << " not installed properly";
```
monich: I believe `QDebug` is inserting spaces between the strings, there's no need to add more.

Venemo: Yep.

dcaliste:
And now, also allowing PGP encryption / decryption:
<pre>
[nemo@Sailfish devel]$ cat data0
Test plaintext data
[nemo@Sailfish devel]$ secrets-tool --encrypt org.sailfishos.crypto.plugin.gnupg.openpgp org.sailfishos.crypto.plugin.gnupg.openpgp "Damien Caliste <dcaliste@free.fr>" F58016BD85212EFF4763BC70E7F32BC76D8C1C73 data0 | tee data0.enc
IV:
LS0tLS1CRUdJTiBQR1AgTUVTU0FHRS0tLS0tClZlcnM=
aW9uOiBHbnVQRyB2Mi4wLjQgKEdOVS9MaW51eCkKCmg=
UUdNQXl0RTlacXJtN1N6QVF2L1VyVWxrNXlXSjMrK2g=
VWFCYnZvTmVkcVNwM3hyVUxRdmtIWjZFdHhMSnFxeAo=
[…]
[nemo@Sailfish devel]$ secrets-tool --decrypt org.sailfishos.crypto.plugin.gnupg.openpgp org.sailfishos.crypto.plugin.gnupg.openpgp "Damien Caliste <dcaliste@free.fr>" F58016BD85212EFF4763BC70E7F32BC76D8C1C73 data0.enc
Test plaintext data

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,277 @@
+/*
+ * Copyright (C) 2018 Damien Caliste.
+ * Contact: Damien Caliste <dcaliste@free.fr>
+ * All rights reserved.
+ * BSD 3-Clause License, see LICENSE.
+ */
```
chriadam: best to define a header guard even for private headers with known inclusion locations.

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,277 @@
...
+    {
+        err = gpgme_engine_check_version(protocol);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            gpgme_engine_info_t info;
+            gpgme_error_t err2;
+            err2 = gpgme_get_engine_info(&info);
+            if (!err2) {
+                while (info) {
+                    if (info->protocol == protocol) {
+                        qWarning() << "protocol:" << gpgme_get_protocol_name(info->protocol);
```
chriadam: instead of qWarning, best to use qCWarning() and define a plugin-specific logging category.  Hrm, or perhaps we need to add a generic secrets daemon plugin logging category, which all plugins should use...

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,277 @@
...
+    }
+    bool operator!()
+    {
+        return (ctx == 0);
+    }
+    operator gpgme_ctx_t() const
+    {
+        return ctx;
+    }
+    QString error()
```
chriadam: can this be const or is either gpgme_error_code(err) or gpgme_strerror(err) nonconst?

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,277 @@
...
+    {
+        err = gpgme_data_new(&data);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR)
+            data = 0;
+    }
+    GPGmeData(const QByteArray &origin)
+        : data(0), err(0)
+    {
+        err = gpgme_data_new_from_mem(&data, origin.constData(), origin.length(), 0);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR)
```
chriadam: Sailfish OS coding style recommends using braces even for single-line conditional expression blocks.  Also elsewhere in this file / PR.

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,277 @@
...
+
+        gpgme_subkey_t subkey = key->subkeys;
+        while (subkey) {
+            if (!strcmp(subkey->fpr, fpr))
+                return true;
+            subkey = subkey->next;
+        }
+        return false;
+    }
+    const char* fpr() const
```
chriadam: what does fpr stand for?

plugins/gnupgplugin/gpgmebase.cpp
```diff
@@ -0,0 +1,761 @@
...
+        return Result(Result::CryptoPluginKeyGenerationError,
+                      QStringLiteral("missing email custom parameter."));
+    }
+
+    GPGmeContext ctx(m_protocol, home);
+    if (!ctx)
+        return Result(Result::CryptoPluginKeyGenerationError, ctx.error());
+
+    QString gnupgKeyParms = "<GnupgKeyParms format=\"internal\">\n";
+    if (kpgParams.keyPairType() == KeyPairGenerationParameters::KeyPairDsa) {
```
chriadam: in this case, you could do: DsaKeyPairGenerationParameters dkpgp(kpgParams); and then read the key-length from the modulusLength() property of the dkpgp instance.

plugins/gnupgplugin/gpgmebase.cpp
```diff
@@ -0,0 +1,761 @@
...
+        gnupgKeyParms += "Key-Type: DSA\n";
+        gnupgKeyParms += "Key-Length: 1024\n";
+        gnupgKeyParms += "Subkey-Type: DSA\n";
+        gnupgKeyParms += "Subkey-Length: 1024\n";
+    } else if (kpgParams.keyPairType() == KeyPairGenerationParameters::KeyPairRsa) {
+        gnupgKeyParms += "Key-Type: RSA\n";
+        gnupgKeyParms += QStringLiteral("Key-Length: %1\n").arg(static_cast<RsaKeyPairGenerationParameters>(kpgParams).modulusLength());
```
chriadam: I'm surprised the static cast works, actually... maybe better to simply in-place construct: RsaKeyPairGenerationParameters(kpgParams).modulusLength();

plugins/gnupgplugin/gpgmebase.cpp
```diff
@@ -0,0 +1,761 @@
...
+    } else {
+        GPGmeData pub, priv;
+        if (!pub || !priv) {
+            return Result(Result::CryptoPluginKeyGenerationError,
+                          QStringLiteral("cannot create data."));
+        }
+ 
+        err = gpgme_op_genkey(ctx, gnupgKeyParms.toUtf8().constData(), pub, priv);
+
+        // Todo do something with the public data.
```
chriadam: is the public data not already used above with the gpgme_op_genkey() call?

plugins/gnupgplugin/gpgmebase.cpp
```diff
@@ -0,0 +1,761 @@
...
+    GPGmeContext ctx(m_protocol);
+    if (!ctx)
+        return Result(Result::StorageError, ctx.error());
+
+    GPGmeKey gkey(ctx, identifier.name(),
+                  (keyComponents & Key::SecretKeyData)
+                  ? GPGmeKey::Secret : GPGmeKey::Public);
+    if (!gkey) {
+        // Todo: need to set the storage error code to
+        // Sailfish::Secrets::Result::InvalidSecretError
```
chriadam: there should be a constructor for this now :-)

plugins/gnupgplugin/gpgmebase.cpp
```diff
@@ -0,0 +1,761 @@
...
+    if (!ctx)
+        return Result(Result::CryptoPluginDecryptionError, ctx.error());
+
+    gpgme_error_t err;
+    err = gpgme_op_decrypt_verify(ctx, gdata, output);
+    if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+        return Result(Result::CryptoPluginDecryptionError,
+                      QStringLiteral("cannot decrypt message: %1.").arg(gpgme_strerror(err)));
+    }
+    Result verifyResult = verify(CryptoManager::OperationDecrypt,
```
chriadam: What does this do?

plugins/gnupgplugin/gpgme_p.h
dcaliste: fingerprint, changed everywhere.

plugins/gnupgplugin/gpgmebase.cpp
dcaliste: In fact, for DSA, GnuPG handle only length of 1024. I've added a warning in case the user would have chosen another value.

plugins/gnupgplugin/gpgmebase.cpp
dcaliste: Updated.

plugins/gnupgplugin/gpgmebase.cpp
dcaliste: I've updated the comment. Here it's to do something with the generated data from gpgme that is a certificate request in S/MIME.

plugins/gnupgplugin/gpgmebase.cpp
dcaliste: I forgot to upgrade this ;)

plugins/gnupgplugin/gpgmebase.cpp
dcaliste: Name of function changed for checkGPGmeVerification() which is hopefully more descriptive.

rpm/sailfish-secrets.spec
```diff
@@ -225,6 +225,9 @@ BuildRequires:  pkgconfig(Qt5DBus)
 BuildRequires:  pkgconfig(dbus-1)
 BuildRequires:  pkgconfig(libcrypto)
 BuildRequires:  qt5-plugin-sqldriver-sqlite
+BuildRequires:  pkgconfig(libgpg-error)
```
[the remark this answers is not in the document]
dcaliste: That's wise. Changed for this.

plugins/gnupgplugin/gpgme_p.h
dcaliste: I've tried to change this everywhere. Hopefully I've not missed a place.

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,277 @@
...
+            return;
+        }
+        err = gpgme_set_protocol(ctx, protocol);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR) {
+            gpgme_release(ctx);
+            ctx = 0;
+            return;
+        }
+        if (home.isEmpty())
+            return;
```
monich: Return just like that - without releasing the context or anything? Or should this check be done earlier, before the context is allocated?

plugins/gnupgplugin/gpgme_p.h
dcaliste: I've added a commit adding logging category for plugins and used it accordingly.

plugins/gnupgplugin/gpgme_p.h
dcaliste: The context will be automatically freed when the structure will become out of scope by the structure destructor.

dcaliste:
I've amended the commits but I still didn't have time to check that it's working and test. Don't approve the PR immediately. I'll do it tomorrow. Thanks for the review.

plugins/gnupgplugin/gpgme_p.h
```diff
@@ -0,0 +1,277 @@
...
+        : data(0), err(0)
+    {
+        err = gpgme_data_new(&data);
+        if (gpgme_err_code(err) != GPG_ERR_NO_ERROR)
+            data = 0;
+    }
+    GPGmeData(const QByteArray &origin)
+        : data(0), err(0)
+    {
+        err = gpgme_data_new_from_mem(&data, origin.constData(), origin.length(), 0);
```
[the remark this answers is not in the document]
dcaliste: In fact, no copy is done, see the 0 argument on last position. Doc in https://www.gnupg.org/documentation/manuals/gpgme/Memory-Based-Data-Buffers.html

I may #define NO_COPY 0 to be clear...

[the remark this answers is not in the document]
monich: Then if it's not copy, then you need to carry QByteArray around too, to make sure that the actual data don't get deallocated before GPGmeData.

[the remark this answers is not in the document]
dcaliste: These are just helpers in gpgme_p.h to automatically have deallocation. I kept the same restrictions than in gpgme API. Callers are responsible to keep the data alive indeed. Since these helpers are not passed between routines I think it's fine.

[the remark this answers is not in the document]
dcaliste: I've added a comment to relmind that origin QByteArray object should be alive for the whole life of the created structure and replaced the last 0 argument with a NO_COPY #define.

Thank you for your comments.

dcaliste:
@chriadam ok, I've quickly tested on device that signing and verifying are working with the secrets-tool CLI, new automatic test for GnuPG is also passing on device. We can proceed.

dcaliste:
Added encryption / decryption tests and rebased.

dcaliste:
I've modified a bit the pinentry code to handle caching and added also a DConf entry to set wether or not the user wants to store passphrases.

I've changed the ```fprintf``` for qCDebug() also.

dcaliste:
@chriadam I've discovered a funny issue: when running daemon from CLI (with a simple `devel-su -p sailfish-secretsd`), the GnuPG plugins were properly loaded. But when launched by systemd, no GnuPG plugins.

It comes from the fact that invoker is currently running the daemon as `type=qt5`, while GnuPG plugins are using non-qt5 libraries (like gpgme). So I've changed the service file to run invoker with `type=generic`, I hope it's not an issue.

rpm/sailfish-secrets.spec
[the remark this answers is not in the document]
chriadam: After more thought, is there any way to implement the pinentry component such that it can be licensed with BSD-3-clause also?  Given that we were hoping to distribute this gnupg plugin along with the daemon, I'd prefer a maximally permissive license to avoid potential inadvertent violations...

chriadam:
Regarding changing it to a generic booster, that's perfectly fine.  Aside from that, also LGTM, although I'm concerned that I don't know about about the implications of a GPLv2 licensed plugin given that we intend to distribute the plugin along with the daemon (or is it ok, given that the daemon is even more permissively licensed?  IANAL...)

dcaliste:
Thank you @chriadam very much for looking at this.

GPGme has two licenses as far as I can see from the two COPYING and COPYING.LESSER in the sources. One of them is LGPL, so there is no issue linking a plugin of a BSD executable with a LGPL library (read GPGme). I think that the plug-in is written with BSD headers, right ?

The pinentry program is another matter since it is a program, interfacing with GnuPG via IPC and linking the BSD sailfish-secret library. So this program can be whatever we want without any issues. I've chosen the GPL, but I can change my mind if there are reasons that I didn't foresee.

rpm/sailfish-secrets.spec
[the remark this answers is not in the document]
dcaliste: I understand, but the pinentry program is not linking with the daemon. In fact, it ideally should be part of GnuPG itself which is GPL, like other pinentries. Indeed, the pinentry program is used by GnuPG executable to ask for user entry and is using IPC (via libassuan which is LGPL) to interface with GnuPG. So it may even be whatever license.

Our pinentry implementation is then linking with libsecret to interface nicely with SFOS environment. But since libsecret is BSD licensed, the pinentry can be whatever restrictive license we want.

So I think putting the pinentry under GPL won't violate anything. If you think differently, I can change it for BSD though.

chriadam:
I'd prefer that each package used just a single license, and pinentry is packaged with the plugin in this case; thus, if possible, I would prefer it to be licensed as BSD also.  Basically, I have an aversion to accepting any code into the repository which has a less permissive license than that of the rest, mostly because I don't entirely understand the ramifications of those different licenses, and what they might mean e.g. for our partners also.

Finally, it simplifies communication with partner vendors; otherwise, if there is a package which has multiple licenses, someone would have to potentially analyse each source file involved in generating the package to determine which parts may or may not be involved, and what the potential implications might be, when making a decision about what risks are implied when generating a device image with the software included.

Of course, it's your contribution, so if you strong prefer to keep pinentry as GPL in this case, I will simply ask Jolla Legal to definitively look into this case to give the green light, and then once that happens I will merge as-is.

chriadam:
A couple of things I noticed after rebasing on top of master and running the tests:
1) you should generate a test plugin (i.e. whose name ends with ".test") similarly to the other plugins, so that we can run the unit test after starting the daemon in --test mode
2) on my device there is a failure in the addSubKey() test:

QDEBUG : tst_gnupgplugin::addSubKey() [D] unknown:0 - "Collection John Doe <john.doe@example.com> does not exist"
FAIL!  : tst_gnupgplugin::addSubKey() Compared values are not the same
   Actual   (dskr.result().code()): Failed
   Expected (Result::Succeeded)   : Succeeded
   Loc: [tst_gnupgplugin.cpp(193)]

Do I need to install some package which I'm missing?

dcaliste:
> on my device there is a failure in the addSubKey() test:

That's expected. It's due to the fact that DeleteStoredKeyRequest is not passing the custom parameters (because DeleteSecretRequest from Secrets don't have it, only requests from Crypto, and the custom parameters are dropped when calling the DeleteSecretRequest from the DeleteStoredKeyRequest). We discussed this a long time ago. I proposed to expand the Secrets plugin API, but I didn't find the time to actually propose a PR for it.

So for the moment, I've disabled this part of the test and put a comment about the API limitation.

dcaliste:
> you should generate a test plugin (i.e. whose name ends with ".test")

Ok, done for the OpenPGP plugin. What can I run to check if the plugin passes the unit tests ? The daemon with `--test` is properly able to load the test plugin but I don't know what to run after. Besides, I'm guessing it will not work out of the box because key generation for OpenPGP is a bit specific and requires some custom parameters besides key name and collection name.

chriadam:
Ok, sounds good.  When running the daemon in --test mode, I just expect to be able to run the tst_gnupgplugin test and have it pass.  If that now works, then LGTM.  (I'll double check that on my device this afternoon.)  The other unit tests should still pass (i.e. they shouldn't fail just because this package is installed) but they needn't actually test any functionality from this plugin.

chriadam:
Thank you very much.  Merging now.

PR #71, opened 2018-04-17
[sailfish-crypto] Implement example USB token for documentation purposes. Contributes to JB#40058

chriadam:


plugins/sqlcipherplugin/cryptoplugin.cpp
```diff
@@ -344,7 +344,13 @@ Sailfish::Secrets::Daemon::Plugins::SqlCipherPlugin::sign(
...
         QByteArray *signature)
 {
-    return m_opensslCryptoPlugin.sign(data, key, padding, digestFunction, customParameters, signature);
+    Sailfish::Crypto::Key fullKey;
+    Sailfish::Crypto::Result keyResult = getFullKey(key, &fullKey);
+    if (keyResult.code() != Sailfish::Crypto::Result::Succeeded) {
+        return keyResult;
+    }
+
+    return m_opensslCryptoPlugin.sign(data, fullKey, padding, digestFunction, customParameters, signature);
```
Venemo: This was already fixed here:

https://github.com/sailfishos/sailfish-secrets/commit/1639eee8fdcbcf7a136a6694f55d60591f42df0a#diff-c7cac238723147f70a648d3efdeaccf8R621

Venemo: Though I agree with the better error handling, but don't see why you removed the fix from the `OpenSslCryptoPlugin`.

jpetrell:
LGTM

plugins/sqlcipherplugin/cryptoplugin.cpp
chriadam: I did it to improve encapsulation - the openssl crypto plugin cannot itself store keys, so it makes no sense for it to try to get the full key.  Instead, the sqlcipher plugin should do that, then pass the full key to the openssl crypto plugin.

It's just cleaner, even though practically speaking it should make little difference.

PR #74, opened 2018-04-20
[sailfish-secrets] Add string translation ids. Contributes to JB#36797

chriadam:


jpetrell:
Key, collection and plugin names probably should be highlighted from the sentences somehow (displayed as secondary highlight color or in quotation marks). I'll check with design.

Just realized key, collection and plugin names can be arbitrarily long. Need to rethink the design (but maybe not in first version as the schedule is tight)

daemon/CryptoImpl/cryptorequestprocessor.cpp
```diff
@@ -446,7 +454,12 @@ Daemon::ApiImpl::RequestProcessor::generateStoredKey_afterPreCheck(
             ikdRequest.setPluginName(keyTemplate.identifier().storagePluginName());
             ikdRequest.setOperation(Sailfish::Secrets::InteractionParameters::DeriveKey);
             ikdRequest.setAuthenticationPluginName(uiParams.authenticationPluginName());
-            ikdRequest.setPromptText(uiParams.promptText());
+            //: This will be displayed to the user, prompting them to enter a passphrase from which a key will be derived.  %1 is the key name, %2 is the collection name, %3 is the plugin name.
+            //% "An application wants to store a new key named %1 within collection %2 in plugin %3.  Enter a passphrase from which the key will be derived."
```
jpetrell: "..in plugin %3.  Enter a pass..." -> ..in plugin %3. Enter a pass..." ? We don't do double spaces after comma in translation strings?

plugins/inappauthplugin/plugin.cpp
```diff
@@ -57,8 +55,8 @@ Daemon::Plugins::InAppPlugin::beginUserInputInteraction(
                                   Q_ARG(Sailfish::Secrets::Result, Sailfish::Secrets::Result(Sailfish::Secrets::Result::Succeeded)),
                                   Q_ARG(QByteArray, QByteArray("example passphrase for unit test")));
         return Result(Result::Pending);
-    } else if (interactionParameters.promptText()
-            == QLatin1String("Enter the old master lock code for device secrets")) {
+    } else if (interactionParameters.promptText().endsWith(
+                   QLatin1String("Enter the old master passphrase."))) {
```
jpetrell: Isn't the prompt text is already localized here?

plugins/inappauthplugin/plugin.cpp
```diff
@@ -100,7 +99,8 @@ Daemon::Plugins::InAppPlugin::beginUserInputInteraction(
                                   Q_ARG(Sailfish::Secrets::Result, Sailfish::Secrets::Result(Sailfish::Secrets::Result::Succeeded)),
                                   Q_ARG(QByteArray, QByteArray()));
         return Result(Result::Pending);
-    } else if (interactionParameters.promptText().startsWith(QLatin1String("Enter the new lock code for the plugin"))) {
+    } else if (interactionParameters.promptText().endsWith(
+                   QLatin1String("Enter the new passphrase for the plugin."))) {
```
jpetrell: Entering new passphrase flows here only require typing password once. There is quite a high likelihood that user mistypes the password so repeat step should be required.

plugins/inappauthplugin/plugin.cpp
```diff
@@ -81,8 +79,8 @@ Daemon::Plugins::InAppPlugin::beginUserInputInteraction(
                                   Q_ARG(Sailfish::Secrets::Result, Sailfish::Secrets::Result(Sailfish::Secrets::Result::Succeeded)),
                                   Q_ARG(QByteArray, newLockCount % 2 == 1 ? QByteArray("masterlock") : QByteArray()));
         return Result(Result::Pending);
-    } else if (interactionParameters.promptText()
-               == QLatin1String("Provide the master unlock code for device secrets")) {
+    } else if (interactionParameters.promptText().endsWith(
+                   QLatin1String("Enter the master passphrase to unlock the secrets service."))) {
```
jpetrell: Need to discuss with design what to call Sailfish OS credentials store.

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -1017,8 +1027,13 @@ Daemon::ApiImpl::RequestProcessor::setCollectionSecretGetAuthenticationCode(
         ikdRequest.setOperation(InteractionParameters::StoreSecret);
         ikdRequest.setInputType(InteractionParameters::AlphaNumericInput);
         ikdRequest.setEchoMode(InteractionParameters::PasswordEchoOnEdit);
-        ikdRequest.setPromptText(tr("Enter the passphrase to unlock the collection %1 in order to store secret %2")
-                                 .arg(secret.identifier().collectionName(), secret.identifier().name()));
+        //: This will be displayed to the user, prompting them to enter the passphrase to unlock the collection in which a new secret will be stored.  %1 is the application name, %2 is the  secret name, %3 is the collection name, %4 is the plugin name.
+        //% "Application %1 wants to store a new secret named %2 into collection %3 in plugin %4.  Enter the passphrase to unlock the collection."
```
jpetrell: In UI design "Enter the passphrase.." strings are shown on separate label, requiring two separate translation strings.

In Sailfish OS we normally talk about passwords and not passphrases.

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -343,8 +343,10 @@ Daemon::ApiImpl::RequestProcessor::createCustomLockCollection(
     ikdRequest.setOperation(InteractionParameters::CreateCollection);
     ikdRequest.setInputType(InteractionParameters::AlphaNumericInput);
     ikdRequest.setEchoMode(InteractionParameters::PasswordEchoOnEdit);
-    ikdRequest.setPromptText(tr("Enter the passphrase which will be used to encrypt the new collection %1 in plugin %2")
-                             .arg(collectionName, storagePluginName));
+    //: This will be displayed to the user, prompting them to enter a passphrase which will be used to encrypt a collection.  %1 is the application name, %2 is the collection name, %3 is the plugin name.
+    //% "Application %1 wants to create a new secrets collection %2 in plugin %3.  Enter the passphrase which will be used to encrypt the collection."
```
jpetrell: UI lingo is "apps", not "applications". If app icon is shown in the future I would drop the whole "App" prefix.

daemon/CryptoImpl/cryptorequestprocessor.cpp
```diff
@@ -446,7 +454,12 @@ Daemon::ApiImpl::RequestProcessor::generateStoredKey_afterPreCheck(
             ikdRequest.setPluginName(keyTemplate.identifier().storagePluginName());
             ikdRequest.setOperation(Sailfish::Secrets::InteractionParameters::DeriveKey);
             ikdRequest.setAuthenticationPluginName(uiParams.authenticationPluginName());
```
jpetrell: To me it is not immediately obvious what idk stands for. Also weird to name variable request, which is not request class.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -72,6 +72,7 @@ void Daemon::ApiImpl::SecretsDBusObject::userInput(
     modifiedParams.setOperation(InteractionParameters::RequestUserData);
     modifiedParams.setCollectionName(QString());
     modifiedParams.setSecretName(QString());
+    modifiedParams.setPluginName(QString());
```
jpetrell: Aren't these empty string setter calls unnecessary?

daemon/CryptoImpl/cryptorequestprocessor.cpp
chriadam: Oops, I will fix all those cases.  Old habits for manual typesetting ;-)

plugins/inappauthplugin/plugin.cpp
chriadam: Yes, but this is only in the ifdef block for unit test builds.  Basically, this allows the plugin to automatically return some "user input" without actually requiring user interaction.  We should extend these cases to return data automatically for every unit test, to avoid the red-rectangle-prompt...

daemon/SecretsImpl/secretsrequestprocessor.cpp
chriadam: I'm not sure if the authentication dialogs Andrew has implemented support multiple separated labels - perhaps they do, but not sure.

I will change passphrase to password.

daemon/CryptoImpl/cryptorequestprocessor.cpp
chriadam: IKD is input key data - we generate a key from the key data via key derivation.  I can rename these to passphraseQuery or something instead.

daemon/SecretsImpl/secrets.cpp
chriadam: Yes, this is the case where we are receiving the interaction parameters from the client application (e.g in case they are specifying echo mode, alphanumeric vs pin-entry, etc) - but we don't want them to set the collection/secret/plugin name (so we overwrite it with empty data).

This affects later on how we generate the prompt text, as when the Crypto side asks the Secrets side to perform the same sort of request (userInput) it will pre-fill the plugin+collection+secretName data, so that we generate the appropriate string, without including a "warning - data will be returned to the application" prefix, etc.

chriadam: I'll add a code comment here to clarify this.

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -648,7 +650,10 @@ Daemon::ApiImpl::RequestProcessor::deleteCollectionWithMetadata(
         ikdRequest.setOperation(InteractionParameters::DeleteCollection);
         ikdRequest.setInputType(InteractionParameters::AlphaNumericInput);
         ikdRequest.setEchoMode(InteractionParameters::PasswordEchoOnEdit);
-        ikdRequest.setPromptText(QObject::tr("sailfish_secrets_delete_collection_input_key_data_prompt"));
+        //: This will be displayed to the user, prompting them to enter a passphrase which will be used to unlock a collection for deletion.  %1 is the application name, %2 is the collection name, %3 is the plugin name.
+        //% "Application %1 wants to delete a secrets collection %2 in plugin %3.  Enter the passphrase which will be used to unlock the collection for deletion."
```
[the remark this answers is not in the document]
chriadam: removed

daemon/SecretsImpl/secretsrequestprocessor.cpp
chriadam: Ok, changed to "App %1 ..."

daemon/CryptoImpl/cryptorequestprocessor.cpp
chriadam: Renamed to promptParams

plugins/inappauthplugin/plugin.cpp
chriadam: Hrm, looking at https://github.com/sailfishos/sailfish-secrets/pull/15#issuecomment-365584951 I'm not sure how to set this, actually.  Will it automatically do it if I provide the repeatNewPasswordText and passwordMismatchText properties?

chriadam: @adenexter which properties do I need to set to make the prompt require the user to re-enter the new password?

jpetrell: Also if user mistypes can we display error and ask user to try again?

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -764,21 +769,23 @@ Daemon::ApiImpl::RequestProcessor::userInput(
...
         // this is a request on behalf of a client application.
         // the user needs to be warned that the data they enter cannot
         // be considered to be "secure" in the secrets-storage sense.
-        const QString warningPromptText = QString::fromLatin1(
-                    "An application is requesting input which will be returned to the application: %1")
-                .arg(ikdRequest.promptText());
-        ikdRequest.setPromptText(warningPromptText);
+
+        //: Inform the user that the application is requesting data. %1 is the application name, %2 is the prompt text supplied by the application.
+        //% "An application %1 is requesting input which will be returned to the application: %2"
```
jpetrell: An application -> An app

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -1009,20 +1019,25 @@ Daemon::ApiImpl::RequestProcessor::setCollectionSecretGetAuthenticationCode(
...
+        InteractionParameters promptParams;
+        promptParams.setApplicationId(callerApplicationId);
+        promptParams.setPluginName(secret.identifier().storagePluginName());
+        promptParams.setCollectionName(secret.identifier().collectionName());
+        promptParams.setSecretName(secret.identifier().name());
+        promptParams.setOperation(InteractionParameters::StoreSecret);
+        promptParams.setInputType(InteractionParameters::AlphaNumericInput);
+        promptParams.setEchoMode(InteractionParameters::PasswordEchoOnEdit);
+        //: This will be displayed to the user, prompting them to enter the passphrase to unlock the collection in which a new secret will be stored. %1 is the application name, %2 is the  secret name, %3 is the collection name, %4 is the plugin name.
+        //% "App %1 wants to store a new secret named %2 into collection %3 in plugin %4. Enter the passphrase to unlock the collection."
```
jpetrell: App %1 -> %1 App

collection %3 -> %3 collection

Like said I would drop "App" when/if app icon can be fetched and shown inside the dialog (would then require two variants of each string).

"plugin" is null-word and quite technical term, I would drop it and just display the name of the plugin.

jpetrell:
Can we limit the length of secret, collection and plugin names to 32 characters? Mention in documentation. Or allow but warn that longer names may get truncated. And truncate with "..." when passing through dialog API.

Also ideally the names (app, secret, collection, plugin) would be highlighted from body text. We don't show quotes in UI, but normally color the highlighted content differently. In this case body text could be shown in secondary highlight color and highlighted names in full highlight color. I.e. Label { color: Theme.secondaryHighlightColor; text: "Secret <font style='color: " + Theme.highlightColor + '">" + secretName + "</font>" }

Earlier mentioned the possibility of plugins to provide icon to represent the plugin. Maybe category enumeration would be better, and OS would provide appropriate Sailfish-style icons (SIM card, USB token, database, generic device icon, etc.). Then we could perhaps also variate the translations to use meaningful names instead of talking about plugins.

daemon/SecretsImpl/secretsrequestprocessor.cpp
jpetrell: So "App %1 wants to store a new secret named %2 into collection %3 in plugin %4." -> "%1 App wants to store a new secret, named %2, into %3 collection from %4."

chriadam: Why reorder that way?  To me "collection X" sounds more natural than "X collection", similarly "App Y" sounds more natural than "Y app".

For the icon stuff, can we postpone that because I'm not sure what we support in the dialogs yet, needs more investigation and potentially dbus API changes etc.

dcaliste:
> I expect fully-qualified-domain-name prefixes, so the plugin names will be long and the "important" information is likely to be RHS

It would be convenient to add in the PluginBase class a displayName() function to return a human readable string naming the plugin. After all, the real name of the plugin is only useful for DBUS internal machinery and users don't care about it (except to call DBUS themselves, …). What do you think ?

chriadam:
As discussed on IRC, it would have to be unique (otherwise malicious plugin could (if installed...) "pretend" to be another storage plugin; or at the very least, a collision could cause confusion in users).  If it is unique, then we may as well use the display name as the real name.  We could refuse to load a plugin if its display name isn't unique, but, the less specific the name is, the more likely a collision is to occur, and then we have plugin-load-ordering issues (e.g. if a new plugin is installed which collides, it might shadow a pre-existing plugin, and if the new plugin is loaded first, then the old plugin becomes unusable / data stored there is "lost").  Need to think about this more, I guess - in practice, maybe it's not a problem...

daemon/SecretsImpl/secretsrequestprocessor.cpp
jpetrell: Yeah let's postpone the icon support.

chriadam:
Ok, I'll integrate this now.

As discussed on IRC, early next week I will add displayName() to PluginBase API, and then modify all of these prompt translation strings to take .arg(plugin(pluginName)->displayName()), instead of just .arg(pluginName).

PR #83, opened 2018-05-01
[sailfish-secrets] Add displayName() to PluginBase API. Contributes to JB#36797

chriadam:
Note that the display name should be informative and human-readable,
however we do not enforce uniqueness (in order to avoid plugin
ordering issues resulting in lost-data issues).

Plugin implementors should take care to ensure that the display name
reported by their plugin is useful and not confusing.

Contributes to JB#36797


chriadam:
Rebased, merging.

PR #85, opened 2018-05-02
[sailfish-secrets] Add default plugin packages that provides plugins. Contributes to JB#40058

rainemak:
This commit adds sailfishsecretsdaemon-secretsplugin-common that
contains plugins are mandatory for all platforms.

In addition, this commit adds four virtual packages
that are used to build proper requirement chain from upper layers.
The lower layers shall depend on the concreate packages that brings
in necessary packages satisfying upper layers.

Provides: sailfishsecretsdaemon-secretsplugins
- All secrets plugins

Provides: sailfishsecretsdaemon-secretsplugin-ssl
- Implements ssl secrets plugin e.i. openssl secrets

Provides: sailfishsecretsdaemon-secretsplugin-sql
- Implements sql secrets plugin e.i. sqlite

Provides: sailfishsecretsdaemon-cryptoplugins
- All secrets plugings

Provides: sailfishsecretsdaemon-cryptoplugin-ssl
- Implements ssl crypto plugin e.i. openssl crypto plugin

### DIFF
--- a/rpm/sailfish-secrets.spec
+++ b/rpm/sailfish-secrets.spec
@@ -159,7 +159,7 @@ Requires:   libsailfishcrypto = %{version}-%{release}
 %{summary}.
 
 %package -n %{secretsdaemon}
-Summary:    Sailfish OS secrets daemon (example).
+Summary:    Sailfish OS secrets daemon.
 Group:      Applications/System
 BuildRequires:  pkgconfig(Qt5Core)
 BuildRequires:  pkgconfig(Qt5DBus)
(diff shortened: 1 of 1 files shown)


chriadam:
One comment is that e.g. the SQLCipher plugin uses OpenSSL for encryption (and offers crypto capabilities).  So should it be installed by sailfishsecretsdaemon-secretsplugin-ssl or sailfishsecretsdaemon-secretsplugin-sql or both?

I'd be more inclined to package every plugin separately, and include just the passwordagent + sqlcipher plugin the default set.

rpm/sailfish-secrets.spec
```diff
@@ -182,9 +182,13 @@ Requires:         qt5-plugin-sqldriver-sqlcipher
 Provides an example secrets storage and cryptographic operations system daemon service,
 which exposes functionality provided by libsailfishsecrets and libsailfishcrypto to clients via DBus.
 
-%package -n %{secretsdaemon}-secretsplugins
+%package -n %{secretsdaemon}-secretsplugins-default
 Summary:    Sailfish OS secrets daemon (example) plugins.
 Group:      Applications/System
+Provides: %{secretsdaemon}-secretsplugins
```
saukko: What should depend on this? %{secretsdaemon}?

rpm/sailfish-secrets.spec
```diff
@@ -182,9 +182,13 @@ Requires:         qt5-plugin-sqldriver-sqlcipher
 Provides an example secrets storage and cryptographic operations system daemon service,
 which exposes functionality provided by libsailfishsecrets and libsailfishcrypto to clients via DBus.
 
-%package -n %{secretsdaemon}-secretsplugins
+%package -n %{secretsdaemon}-secretsplugins-default
 Summary:    Sailfish OS secrets daemon (example) plugins.
```
saukko: Can we drop the (example) already?

rpm/sailfish-secrets.spec
```diff
@@ -194,12 +198,14 @@ Requires:   qt5-plugin-sqldriver-sqlcipher
 Requires:   %{secretsdaemon} = %{version}-%{release}
 Requires:   libsailfishsecretspluginapi = %{version}-%{release}
 
-%description -n %{secretsdaemon}-secretsplugins
+%description -n %{secretsdaemon}-secretsplugins-default
 %{summary}.
 
-%package -n %{secretsdaemon}-cryptoplugins
+%package -n %{secretsdaemon}-cryptoplugins-default
 Summary:    Sailfish OS crypto daemon (example) plugins.
```
saukko: As above...

PR #87, opened 2018-05-03
[sailfish-secrets] Add more status reporting API to PluginBase. Contributes to JB#41552

chriadam:


chriadam:
Just remembered that I need to include the plugin display name in the PluginInfo struct also.  TODO!

chriadam:
Updated, the displayName is now reported also via PluginInfo.  This is now ready for review.

lib/Crypto/plugininfo.h
```diff
@@ -22,28 +22,36 @@ class PluginInfoPrivate;
 class SAILFISH_CRYPTO_API PluginInfo
 {
     Q_GADGET
+    Q_PROPERTY(QString displayName READ displayName WRITE setDisplayName)
```
dcaliste: Not really related to this PR, but seeing displayName, name, version… properties being writable is kind of strange to me. PluginInfo is returning static information on the plugin.

Looking at the code, the setDisplayName() method and friends are used in the serialization code only if I'm right. Would it be cleaner to make these attributes construct only, or at least private and the serialization class being friend?

What do you think?

Venemo: I don't think it really matters, since this is a simple data transfer object.

chriadam: You're correct that these should be used for reporting only.  However, I'm not certain that limiting the API in that manner is necessary.  For Q_GADGET/value types in general, they can be considered similarly to a struct, where the fields can be overwritten but that has no "meaning" (that is, there's no SetPluginInfoRequest or similar, for good reason - the plugin info isn't writable).

dcaliste: I see the point of having these as light structures potentially exportable to QML, but I like to have APIs that reflect the usecases, when a struct is for passing information, its purpose is clearer with read-only APIs. Anyway, it was just a side remark that can be ignored ;)

chriadam: You are correct, actually.  We should take a note of this case and also other cases in the API which can be improved, and do them in a "improve API hygiene" commit.  Are there any other similar places where you see such room for improvement?

dcaliste: I don't know somewhere else, it struck me yesterday when looking at the PR diff to see a WRITE part in the properties for something that reflects static data.

I'll open a issue, as suggested by @Venemo for another remark, to keep track of these discussions. I agree that they are not of the highest importance and not related to the actual PR.

PR #88, opened 2018-05-03
[sailfish-crypto] Improve ImportKeyRequest API. Contributes to JB#40058

chriadam:
Instead of taking a key parameter, it now takes raw byte array, which
will be converted into an appropriate, usable Key by the plugin.

### DIFF
--- a/lib/CryptoPluginApi/extensionplugins.h
+++ b/lib/CryptoPluginApi/extensionplugins.h
@@ -84,13 +84,14 @@ public:
             Sailfish::Crypto::Key *keyMetadata) = 0;
 
     virtual Sailfish::Crypto::Result importKey(
-            const Sailfish::Crypto::Key &key,
+            const QByteArray &data,
             const QByteArray &passphrase,
             const QVariantMap &customParameters,
             Sailfish::Crypto::Key *importedKey) = 0;
 
     virtual Sailfish::Crypto::Result importAndStoreKey(
-            const Sailfish::Crypto::Key &key,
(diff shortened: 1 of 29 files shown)


lib/CryptoPluginApi/extensionplugins.h
```diff
@@ -84,13 +84,14 @@ class SAILFISH_CRYPTO_API CryptoPlugin : public virtual Sailfish::Secrets::Plugi
             Sailfish::Crypto::Key *keyMetadata) = 0;
 
     virtual Sailfish::Crypto::Result importKey(
-            const Sailfish::Crypto::Key &key,
+            const QByteArray &data,
             const QByteArray &passphrase,
```
dcaliste: Great move, indeed to import keys from file dump, or from the web on key servers. It seems to me that the plugin documentation update is lacking. Especially, in the StoredKey variant what is expected to be in keyTemplate. I mean, what to do if data already contains key name, key operations…

chriadam: Can you expand upon this - do you mean that the key data in the QByteArray can contain metadata like a name, and allowable operations?  Good point - I hadn't thought about that case.  Do you have suggestions for what appropriate requirements should be?  My first thought is that the plugin should store the "most restrictive" set of operations which are defined in either the keyTemplate or the input data, but should always use the identifier specified in the keyTemplate (and otherwise, could potentially return an error instead of storing the key with a mismatched name, perhaps).  What do you think?

PR #89, opened 2018-05-04
[sailfish-crypto] Implement Sign/Verify CipherRequest support. Contributes to JB#40058

chriadam:


plugins/opensslcryptoplugin/opensslcryptoplugin.cpp
```diff
@@ -755,11 +755,12 @@ Daemon::Plugins::OpenSslCryptoPlugin::verify(
         return Sailfish::Crypto::Result(Sailfish::Crypto::Result::Succeeded);
     } else if (r == 0) {
         // Verification performed without error, but signature didn't match.
+        *verified = false;
```
dcaliste: Isn't it better to put *verified = false at the beginning of the routine just after the pointer check, so like that the output argument is always set, and change to true only in the particular case when it makes sense? It makes the code more robust, if a negligent user is using ```verified``` without properly checking the result success, he will be safe.

PR #94, opened 2018-05-07
Expose PluginInfos to QML

rainemak:


qml/Crypto/main.cpp
```diff
@@ -27,6 +27,7 @@ void Sailfish::Crypto::Plugin::CryptoPlugin::registerTypes(const char *uri)
 
     qmlRegisterUncreatableType<Sailfish::Crypto::Request>(uri, 1, 0, "CryptoRequest", QStringLiteral("Request is an abstract class, can't construct in QML"));
     qRegisterMetaType<Sailfish::Crypto::Request::Status>("CryptoRequestStatus");
+    qmlRegisterUncreatableType<Sailfish::Crypto::PluginInfo>(uri, 1, 0, "PluginInfo", QStringLiteral("PluginInfo objects cannot be constructed directly in QML"));
```
Venemo: To avoid the warning about name collision, please change this to `CryptoPluginInfo`:

```
qmlRegisterUncreatableType<Sailfish::Crypto::PluginInfo>(uri, 1, 0, "CryptoPluginInfo", QStringLiteral("PluginInfo objects cannot be constructed directly in QML"));
```

qml/Secrets/main.cpp
```diff
@@ -38,6 +38,7 @@ void Sailfish::Secrets::Plugin::SecretsPlugin::registerTypes(const char *uri)
 
     qmlRegisterUncreatableType<Sailfish::Secrets::Request>(uri, 1, 0, "SecretsRequest", QStringLiteral("Request is an abstract class, can't construct in QML"));
     qRegisterMetaType<Sailfish::Secrets::Request::Status>("SecretsRequestStatus");
+    qmlRegisterUncreatableType<Sailfish::Secrets::PluginInfo>(uri, 1, 0, "PluginInfo", QStringLiteral("PluginInfo objects cannot be constructed directly in QML"));
```
Venemo: To avoid name collision:

```
qmlRegisterUncreatableType<Sailfish::Secrets::PluginInfo>(uri, 1, 0, "SecretsPluginInfo", QStringLiteral("PluginInfo objects cannot be constructed directly in QML"));
```

qml/Crypto/main.cpp
chriadam: Why?  We don't do this for other types (like Result) so why for this?  The client can import the namespace with a qualifier (using the import as syntax) to avoid collision if required.

qml/Secrets/main.cpp
chriadam: Again, what's the reason for this?

qml/Crypto/main.cpp
rainemak: I'd also prefer importing namespace with  a qualifier.

Venemo: @chriadam Actually we do this for other types. Just having an app with two QML import statements should not result in a warning. It is very annoying to a 3rd party developer that he gets a warning for simply importing our stuff.

chriadam: No, we do it for no other imports that I'm aware of.  E.g. we don't prefix any type in the Silica import with Silica, nor from any other QML import namespace we define.  I'm also certain that we discussed this very issue in the past, and the fact that you can import a QML namespace with a qualifier to allow type namespacing to avoid collisions was discussed then also.

PR #95, opened 2018-05-08
Split passphrase prompt text into multiple strings.

adenexter:
Typically this will be a message describing the reason for the prompt
and an 'instruction' describing the what to enter.

For entering new passphrases there are separate 'newInstruction'
and 'repeatInstruction' prompts for the entry with confirmation,
the 'instruction' prompt in reserved for entering an existing
passphrase in the event the prompt is for changing a passphrase.


chriadam:
LGTM

chriadam:
Tested on device and works nicely, although with longer application and collection names, the (automatically opened) VKB sometimes overlaps over some of the label text (especially the second label is sometimes obscured).  I don't think this is a problem which should prevent this from being merged, however, but we might look at improving the UX somehow in the future.

PR #97, opened 2018-05-08
[sailfish-crypto] Use VerificationResult instead of bool for verified…

blammit:
… values. Contributes to JB#40058

Add CryptoManager::VerificationResult for more detailed verification
result codes and use this in the library and plugins instead of bool
values.

### DIFF
--- a/lib/Crypto/cryptomanager.h
+++ b/lib/Crypto/cryptomanager.h
@@ -313,6 +313,20 @@ public:
     Q_DECLARE_FLAGS(Operations, Operation)
     Q_FLAG(Operations)
 
+    enum VerificationStatusType {
+        VerificationStatusUnknown = 0,
+        VerificationSucceeded = 1,
+        VerificationFailed = 2,
+        VerificationSignatureInvalid = 4,
+        VerificationSignatureExpired = 8,
+        VerificationKeyExpired = 16,
+        VerificationKeyRevoked = 32,
+        VerificationKeyInvalid = 64
+    };
+    Q_ENUM(VerificationStatusType)
+    Q_DECLARE_FLAGS(VerificationStatus, VerificationStatusType)
+    Q_FLAG(VerificationStatus)
+
     CryptoManager(QObject *parent = Q_NULLPTR);
     virtual ~CryptoManager();
 
@@ -352,6 +366,9 @@ Q_DECLARE_METATYPE(Sailfish::Crypto::CryptoManager::SignaturePadding);
 Q_DECLARE_METATYPE(Sailfish::Crypto::CryptoManager::DigestFunction);
 Q_DECLARE_METATYPE(Sailfish::Crypto::CryptoManager::Operation);
 Q_DECLARE_METATYPE(Sailfish::Crypto::CryptoManager::Operations);
+Q_DECLARE_METATYPE(Sailfish::Crypto::CryptoManager::VerificationStatusType);
+Q_DECLARE_METATYPE(Sailfish::Crypto::CryptoManager::VerificationStatus);
 Q_DECLARE_OPERATORS_FOR_FLAGS(Sailfish::Crypto::CryptoManager::Operations);
+Q_DECLARE_OPERATORS_FOR_FLAGS(Sailfish::Crypto::CryptoManager::VerificationStatus);
 
 #endif // LIBSAILFISHCRYPTO_CRYPTOMANAGER_H
--- a/lib/Crypto/verifyrequest.h
+++ b/lib/Crypto/verifyrequest.h
@@ -33,7 +33,7 @@ class SAILFISH_CRYPTO_API VerifyRequest : public Sailfish::Crypto::Request
     Q_PROPERTY(Sailfish::Crypto::CryptoManager::SignaturePadding padding READ padding WRITE setPadding NOTIFY paddingChanged)
     Q_PROPERTY(Sailfish::Crypto::CryptoManager::DigestFunction digestFunction READ digestFunction WRITE setDigestFunction NOTIFY digestFunctionChanged)
     Q_PROPERTY(QString cryptoPluginName READ cryptoPluginName WRITE setCryptoPluginName NOTIFY cryptoPluginNameChanged)
-    Q_PROPERTY(bool verified READ verified NOTIFY verifiedChanged)
+    Q_PROPERTY(Sailfish::Crypto::CryptoManager::VerificationStatus verificationStatus READ verificationStatus NOTIFY verificationStatusChanged)
 
 public:
     VerifyRequest(QObject *parent = Q_NULLPTR);
@@ -57,7 +57,7 @@ public:
     QString cryptoPluginName() const;
     void setCryptoPluginName(const QString &pluginName);
 
-    bool verified() const;
+    Sailfish::Crypto::CryptoManager::VerificationStatus verificationStatus() const;
 
     Sailfish::Crypto::Request::Status status() const Q_DECL_OVERRIDE;
     Sailfish::Crypto::Result result() const Q_DECL_OVERRIDE;
@@ -78,7 +78,7 @@ Q_SIGNALS:
     void paddingChanged();
     void digestFunctionChanged();
     void cryptoPluginNameChanged();
-    void verifiedChanged();
+    void verificationStatusChanged();
 
 private:
     QScopedPointer<VerifyRequestPrivate> const d_ptr;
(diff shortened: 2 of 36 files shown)


lib/Crypto/cryptomanager.h
```diff
@@ -313,6 +313,19 @@ class SAILFISH_CRYPTO_API CryptoManager : public QObject
     Q_DECLARE_FLAGS(Operations, Operation)
     Q_FLAG(Operations)
 
+    enum VerificationResult {
+        VerificationSucceeded,
+        VerificationFailed,
+        VerificationInvalidSignature,
+        VerificationValidSignatureExpired,
```
chriadam: can a signature expire?

lib/Crypto/cryptomanager.h
```diff
@@ -313,6 +313,19 @@ class SAILFISH_CRYPTO_API CryptoManager : public QObject
...
+        VerificationValidSignatureWithExpiredKey,
+        VerificationValidSignatureWithRevokedKey,
+        VerificationInvalidKey,
+
+        VerificationResultUnknown = 255
```
chriadam: For all other enums, Unknown had value 0.  I guess it's not important, but is there any reason why a different value was chosen for this one?

lib/Crypto/cryptomanager.h
```diff
@@ -313,6 +313,19 @@ class SAILFISH_CRYPTO_API CryptoManager : public QObject
     Q_DECLARE_FLAGS(Operations, Operation)
     Q_FLAG(Operations)
 
+    enum VerificationResult {
```
chriadam: How about just Verification instead of VerificationResult?  I just don't like shadowing the term Result.  Does anyone have strong feelings for or against?

lib/Crypto/verifyrequest.h
```diff
@@ -33,7 +33,7 @@ class SAILFISH_CRYPTO_API VerifyRequest : public Sailfish::Crypto::Request
     Q_PROPERTY(Sailfish::Crypto::CryptoManager::SignaturePadding padding READ padding WRITE setPadding NOTIFY paddingChanged)
     Q_PROPERTY(Sailfish::Crypto::CryptoManager::DigestFunction digestFunction READ digestFunction WRITE setDigestFunction NOTIFY digestFunctionChanged)
     Q_PROPERTY(QString cryptoPluginName READ cryptoPluginName WRITE setCryptoPluginName NOTIFY cryptoPluginNameChanged)
-    Q_PROPERTY(bool verified READ verified NOTIFY verifiedChanged)
+    Q_PROPERTY(Sailfish::Crypto::CryptoManager::VerificationResult verified READ verified NOTIFY verifiedChanged)
```
chriadam: I'd suggest changing the name of this property to verification, and the getter/notif to verification()/verificationChanged()

chriadam:
A couple of comments, mostly LGTM

lib/Crypto/cryptomanager.h
rainemak: If certificate expires then I'd assume that signature should expire as well?

lib/Crypto/cryptomanager.h
rainemak: No strong feelings for or against but somehow Verification tastes better.

lib/Crypto/verifyrequest.h
rainemak: +1

lib/Crypto/cryptomanager.h
Venemo: How is a signature related to a ceriticate?

lib/Crypto/cryptomanager.h
Venemo: +1

Venemo:
> how to get metadata from signature, preferably in one pass with verification ? Metadata like signer fingerprint.

@dcaliste Is this question related to this PR?

lib/Crypto/cryptomanager.h
dcaliste: In S/MIME protocol, signatures are validated by certificates for instance.

dcaliste:
@Venemo, this is an open question since you plan to break the API with this PR. Do you plan to provide methods to handle metadata, in one pass, with or without breaking the API ? It's something that need to be plan in advance IMHO.

lib/Crypto/cryptomanager.h
blammit: Just "verification" seems to indicate a mode or type, when this is referring to a result. How about VerificationStatus? (I think that mentioned another time actually)

lib/Crypto/verifyrequest.h
blammit: Sounds good

lib/Crypto/cryptomanager.h
blammit: Signatures can be valid but expired according to  https://www.gnupg.org/documentation/manuals/gpgme/Verify.html GPG_ERR_SIG_EXPIRED

lib/Crypto/cryptomanager.h
blammit: I wrote it thinking of Qt APIs which generally have NoError/Success = 0. Though, a lot of them have unknown/generic-type error = 1, rather than = 255. The thing is that VerificationResultUnknown doubles as a busy / result-unknown value -- not sure if we want that, or whether it should always be the generic "VerificationFailed" even in the busy case.

lib/Crypto/cryptomanager.h
chriadam: VerificationStatus sounds good to me

lib/Crypto/verifyrequest.h
chriadam: or verificationStatus

chriadam:
I had forgotten about that.  Is that something that could be provided via a completely separate request (e.g. SignatureMetadataRequest) or does it need to be included in the current VerifyRequest (and CipherRequest) API as another outparameter (QVariantMap signatureMetadata)?

chriadam:
Note that we don't currently deal with PKCS#7 SignedData containers, for example.  The output of our SignRequest is what would become the ContentInfo in a PKCS#7 SignedData message, I believe.  So the other fields (e.g. SignerInfo) is something that we don't deal with currently...

chriadam:
We currently don't support certificates (or certificate chains) in our API currently, so I'm not sure that we can yet support high level concepts which PKCS#7 messages rely on.  Our current SignRequest/VerifyRequest API is much lower-level in that it deals just with QByteArray inputs and outputs.  While it's true that the plugin can require that to be a serialised message format, for example, I'm not sure how much sense it makes to try to support all of those things in the current iteration of the API.

(One quick example: if we added a `QVariantMap signatureMetadata` outparameter to the VerifyRequest, does that mean that we'd also need to add a `QVariantMap signatureMetadata` input parameter to the SignRequest API, which could potentially include some signer information / fingerprint information, but whose values the plugin could override (and add more e.g. the appropriate hash function etc) or is that not required?)

lib/Crypto/cryptomanager.h
chriadam: Our API currently doesn't deal with certificates in any way, as we didn't have time to mature that set of features.  Well, I guess that someone could define a signature format which included some metadata like expiry date, separately from certificates, so we can leave this here for now anyway.

blammit:
PR updated as per discussion.

chriadam:
LGTM, but let's wait a day to see if anyone has any feedback on the discussion regarding other metadata (like SignerInfo etc) above before we merge this one.

dcaliste:
About the PKCS#7 signature and verification handling, GnuPG can do it via the same API than the one we have already. Gpgme library is abstracting for us the identification of certificates like we do for keys. I can show it through the GnuPG plugin next week.

dcaliste:
https://www.gnupg.org/documentation/manuals/gpgme/Verify.html and specifically the ```gpgme_signature_t```. In fact in gpgme the metadata are extracted, saved in the context and retrieved by a later call. But since we don't have context, we cannot do the same.

The metadata are not necessary for signing though because they are already provided via the key argument and possibly the customParameter.

dcaliste:
Sorry, you should read 'not block this PR' in a previous comment. Cannot modify on the phone.

blammit:
I found an issue with the tst_crypotrequest signVerify autotest so I'll fix that first; looking into it now.

chriadam:
LGTM.  Let's get this one in and then consider how to properly support signature metadata in a separate PR.

PR #98, opened 2018-05-08
Refactor signing and verify

Venemo:
This PR is a small refactor of the signing and verification code that removes code duplication and ensures that the same code path is used for both cipher sessions and normal operations, thus reducing the chance of regressions.

### DIFF
--- a/plugins/opensslcryptoplugin/evp/evp.cpp
+++ b/plugins/opensslcryptoplugin/evp/evp.cpp
@@ -1068,22 +1068,98 @@ int OpenSslEvp::verify(const EVP_MD *digestFunc,
                        const uint8_t *signature,
                        size_t signatureLength)
 {
+    EVP_MD_CTX *mdctx = nullptr;
+    int r = verify_session_init(&mdctx, digestFunc, pkey);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "Failed to initialize verify session.", err_dontfree);
+
+    r = verify_session_update(mdctx, bytes, bytesCount);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1, "Failed to update verify session.", err_dontfree);
+
+    r = verify_session_finalize(mdctx, signature, signatureLength);
+    OSSLEVP_HANDLE_ERR(r < 0,, "Failed to finalize verify session.", err_dontfree);
+
+    err_dontfree:
+    return r;
+}
+
+/*
+    int verify_session_init(EVP_MD_CTX **ctx,
+                            const EVP_MD *digestFunc,
+                            EVP_PKEY *pkey);
+*/
+
+int OpenSslEvp::verify_session_init(EVP_MD_CTX **ctx,
+                        const EVP_MD *digestFunc,
+                        EVP_PKEY *pkey)
+{
+    if (ctx == nullptr) {
+        return -2;
+    }
+
+    *ctx = nullptr;
+
     int r = -1;
     EVP_MD_CTX *mdctx = EVP_MD_CTX_create();
-    OSSLEVP_HANDLE_ERR(mdctx == NULL, r = -1, "failed to allocate memory for MD context", err_dontfree);
+    OSSLEVP_HANDLE_ERR(mdctx == nullptr, r = -1, "failed to allocate memory for MD context", err_dontfree);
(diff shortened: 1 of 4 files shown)


plugins/opensslcryptoplugin/evp/evp.cpp
```diff
@@ -848,28 +848,147 @@ int OpenSslEvp::sign(const EVP_MD *digestFunc,
...
-    OSSLEVP_HANDLE_ERR(*signature == NULL, r = -1, "failed to allocate memory for signature", err_free_mdctx);
+    OSSLEVP_HANDLE_ERR(*signature == nullptr, r = -1, "failed to allocate memory for signature", err_free_mdctx);
 
     r = EVP_DigestSignFinal(mdctx, *signature, signatureLength);
-    OSSLEVP_HANDLE_ERR(r != 1, r = -1; OPENSSL_free(*signature), "failed to finalize DigestSign (2nd call)", err_free_mdctx);
+    OSSLEVP_HANDLE_ERR(r != 1, r = -1; OPENSSL_free(*signature), "failed to finalize DigestSign (2nd call)", err_free_signature);
+
+    // Destroy the mdctx, then return the actual result
+    goto err_free_mdctx;
+    return r;
```
chriadam: isn't this return statement dead-code, as it will be bypassed by the goto above?

plugins/opensslcryptoplugin/evp/evp_helpers_p.h
```diff
@@ -60,9 +60,8 @@ struct CipherSessionDataDeleter
         if (csd->evp_cipher_ctx) {
             EVP_CIPHER_CTX_free(csd->evp_cipher_ctx);
         }
-        if (csd->evp_md_ctx) {
-            EVP_MD_CTX_destroy(csd->evp_md_ctx);
-        }
+        // NOTE: csd->evp_md_ctx is already destroyed,
+        //       the _finalize functions will destroy it normally, or the _update ones too if there are errors.
```
chriadam: What about the case where the session is destroyed due to timeout, rather than a call to finalize?

Venemo: You are right, I ought to fix this.

plugins/opensslcryptoplugin/evp/evp.cpp
Venemo: Yes, it's very dead. Surprised the compiler didn't warn about it.

Venemo:
@chriadam Fixed the issues you raised.

Venemo:
@chriadam Rebased on top of latest master and fixed the issue you found. It was tricky but at the end as usual it turned out to be just a one-liner in the refactored verification code. Turns out that I forgot to run `tst_evp` but after it occoured to me it made it quite easy to track down.

plugins/opensslcryptoplugin/evp/evp_helpers_p.h
```diff
@@ -411,3 +411,32 @@ int getEllipticCurveKeySize(Sailfish::Crypto::CryptoManager::EllipticCurve curve
...
+    }
+
+    // Read the public key data into an EVP_PKEY
+    EVP_PKEY *pkeyPtr = Q_NULLPTR;
+    read(bio.data(), &pkeyPtr, Q_NULLPTR, Q_NULLPTR);
+
+    return pkeyPtr;
+}
+
+EVP_PKEY *readEvpPrivKey(const QByteArray privKey)
```
chriadam: should this be a const ref instead of value?  Doesn't matter too much due to CoW semantics, but for consistency

plugins/opensslcryptoplugin/evp/evp_helpers_p.h
```diff
@@ -411,3 +411,32 @@ int getEllipticCurveKeySize(Sailfish::Crypto::CryptoManager::EllipticCurve curve
...
+{
+    return readEvpKey(privKey, PEM_read_bio_PrivateKey);
+}
+
+EVP_PKEY *readEvpPubKey(const QByteArray pubKey)
```
chriadam: ditto

chriadam:
Added const ref for consistency, rebased against master.  LGTM, merging, thanks!

PR #99, opened 2018-05-09
[sailfish-crypto] Rename some Result enum values. Contributes to JB#4…

blammit:
…0058

End error values with 'Error' and make use e.g. OperationNotSupportedError
instead of UnsupportedOperation to make Result values consistent between
Result.h in secrets and crypto.


chriadam:
LGTM too

PR #101, opened 2018-05-10
[sailfish-secrets] Use appropriate auth plugin depending on permissions. Contributes to JB#36797

chriadam:
If a collection is created by a privileged application with its
authentication plugin set to the in-app authentication plugin
(e.g. for in-process flows in Settings application), we don't want
non-privileged third-party applications to be able to use in-app
authentication to unlock the collection (as that would leak the
unlock code to the third-party application).

This commit ensures that we always use the appropriate (system default)
authentication plugin for non-privileged applications, if they do
not own the collection (or standalone secret) in question.

Contributes to JB#36797

### DIFF
--- a/daemon/SecretsImpl/secretsrequestprocessor.cpp
+++ b/daemon/SecretsImpl/secretsrequestprocessor.cpp
@@ -42,6 +42,23 @@ namespace {
                              ident.name()).toUtf8(),
                     QCryptographicHash::Sha512).toBase64());
     }
+
+    QString determineAuthPlugin(const QString &ownerApplicationId,
+                                const QString &callerApplicationId,
+                                bool callerIsPlatformApplication,
+                                const QString &authPluginName,
+                                const QString &interactionServiceAddress,
+                                bool autotestMode) {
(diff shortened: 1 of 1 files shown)


daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -42,6 +42,23 @@ namespace {
...
                     QCryptographicHash::Sha512).toBase64());
     }
+
+    QString determineAuthPlugin(const QString &ownerApplicationId,
+                                const QString &callerApplicationId,
+                                bool callerIsPlatformApplication,
+                                const QString &authPluginName,
+                                const QString &interactionServiceAddress,
+                                bool autotestMode) {
+        const QString defaultPluginName = autotestMode
```
Venemo: In test mode, shouldn't we always just use the in-app plugin?

chriadam: No, some tests explicitly use the password auth plugin (e.g. the test which uses the example usb plugin, and triggers plugin unlocking, etc).

PR #103, opened 2018-05-15
[sailfish-secrets] Add QueryLockStatus operation to LockCodeRequest. Contributes to JB#36797

chriadam:


chriadam:
Ready for review.

daemon/CryptoImpl/crypto.cpp
```diff
@@ -1750,6 +1801,25 @@ void Daemon::ApiImpl::CryptoRequestQueue::handleFinishedRequest(
...
+                // shouldn't happen!
+                qCWarning(lcSailfishCryptoDaemon) << "QueryLockStatusRequest:" << request->requestId << "finished as pending!";
+                *completed = true;
+            } else {
+                bool locked = request->outParams.size()
+                        ? request->outParams.takeFirst().value<bool>()
+                        : true;
+                request->connection.send(request->message.createReply() << QVariant::fromValue<Result>(result)
+                                                                        << QVariant::fromValue<bool>(locked));
+                *completed = true;
```
rainemak: Both branches have *completed = true could be set before break.

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -18,6 +18,28 @@ PluginState Daemon::ApiImpl::pluginState(PluginBase *plugin)
...
+FoundLockedResult Daemon::ApiImpl::queryLockSpecificPlugin(
+        const QMap<QString, Sailfish::Secrets::EncryptionPlugin*> &encryptionPlugins,
+        const QMap<QString, StoragePluginWrapper*> &storagePlugins,
+        const QMap<QString, EncryptedStoragePluginWrapper*> &encryptedStoragePlugins,
+        const QString &lockCodeTarget)
+{
+    bool found = true;
+    bool locked = true;
+    Result result(Result::Succeeded);
+    if (PluginBase *p = storagePlugins.value(lockCodeTarget)) {
```
rainemak: Order ways defined and only one can be target then could be like:
```
    if (PluginBase *p = storagePlugins.value(lockCodeTarget) ||
            p = encryptedStoragePlugins.value(lockCodeTarget) ||
            p = encryptionPlugins.value(lockCodeTarget)) {
        locked = p->supportsLocking() && p->isLocked();
    } else {
        found = false;
    }
```

daemon/SecretsImpl/pluginfunctionwrappers_p.h
```diff
@@ -99,6 +99,16 @@ struct FoundResult {
     Sailfish::Secrets::Result result;
 };
 
+struct FoundLockedResult {
+    FoundLockedResult(bool f = false, bool l = false, const Sailfish::Secrets::Result &r = Sailfish::Secrets::Result())
```
rainemak: bool found = false, bool locked = false, const Sailfish::Secrets::Result &result = Sailfish::Secrets::Result() should not cause issues here with the struct members.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -2030,6 +2110,29 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
...
+                bool locked = request->outParams.size()
+                        ? request->outParams.takeFirst().value<bool>()
+                        : true;
+                if (request->isSecretsCryptoRequest) {
+                    asynchronousCryptoRequestCompleted(request->cryptoRequestId, result, QVariantList() << locked);
+                } else {
+                    request->connection.send(request->message.createReply() << QVariant::fromValue<Result>(result)
+                                                                            << QVariant::fromValue<bool>(locked));
+                }
+                *completed = true;
```
rainemak: Both branches have *completed = true => could be moved before break

daemon/SecretsImpl/secretscryptohelpers.cpp
```diff
@@ -431,6 +458,11 @@ Daemon::ApiImpl::SecretsRequestQueue::asynchronousCryptoRequestCompleted(
             emit userInputCompleted(cryptoRequestId, result, input);
             break;
         }
+        case QueryLockStatusCryptoApiHelperRequest: {
+            bool locked = parameters.size() ? parameters.first().value<bool>() : true;
```
rainemak: I'd have preferred !parameters.isEmpty() or parameters.size() > 0 in favour of implicitly casting size() to bool. This works as well. Just a note.

lib/Crypto/lockcoderequest.h
```diff
@@ -44,6 +46,13 @@ class SAILFISH_CRYPTO_API LockCodeRequest : public Sailfish::Crypto::Request
     };
     Q_ENUM(LockCodeTargetType)
 
+    enum LockStatus {
+        Unknown = 0,
```
rainemak: Mentioned already in irc but I do think that Unsupported or such would be good here if extension plugin does not support locking.

daemon/SecretsImpl/pluginfunctionwrappers.cpp
chriadam: good point, will update to this form

daemon/SecretsImpl/pluginfunctionwrappers_p.h
chriadam: yes, but internal struct, and it's fairly self documenting, will leave like this for consistency with the rest of concurrent function wrapper result structs

daemon/SecretsImpl/secrets.cpp
chriadam: will fix

lib/Crypto/lockcoderequest.h
chriadam: Yep, will add that, thanks

daemon/CryptoImpl/crypto.cpp
chriadam: will fix

rainemak:
LGTM

daemon/SecretsImpl/pluginfunctionwrappers.cpp
```diff
@@ -18,6 +18,33 @@ PluginState Daemon::ApiImpl::pluginState(PluginBase *plugin)
...
+FoundLockStatusResult Daemon::ApiImpl::queryLockSpecificPlugin(
+        const QMap<QString, Sailfish::Secrets::EncryptionPlugin*> &encryptionPlugins,
+        const QMap<QString, StoragePluginWrapper*> &storagePlugins,
+        const QMap<QString, EncryptedStoragePluginWrapper*> &encryptedStoragePlugins,
+        const QString &lockCodeTarget)
+{
+    bool found = true;
+    Sailfish::Secrets::LockCodeRequest::LockStatus lockStatus = Sailfish::Secrets::LockCodeRequest::Unknown;
+    Result result(Result::Succeeded);
+    PluginBase *p = storagePlugins.contains(lockCodeTarget) ? static_cast<PluginBase*>(storagePlugins.value(lockCodeTarget))
```
rainemak: This is a bit unfortunate contains + value. Passing default value to the value simplifies this and avoid extra lookup.

E.g. static_cast<PluginBase*>(storagePlugins.value(lockCodeTarget, nullptr))

chriadam: In this ternary case I don't think it does simplify, as we need to change the assignment if the value is null (i.e. if the map does not contain the value).  I could use chained ifs to get the same behaviour with better performance, so I will change it to that.

chriadam:
Rebased, merging, thanks.

PR #104, opened 2018-05-15
[sailfish-crypto] Add import-stored-key support to test tool. Contributes to JB#41865

chriadam:
Also add links to the ImportKey and ImportStoredKey requests where
appropriate in the documentation.


tools/secrets-tool/commandhelper.cpp
```diff
@@ -467,6 +468,54 @@ void CommandHelper::start(const QString &command, const QStringList &args)
...
+        }
+
+        QFile importFile(importDataFile);
+        if (!importFile.open(QIODevice::ReadOnly)) {
+            qInfo() << "Unable to open import data file for reading!";
+            emitFinished(EXITCODE_FAILED);
+            return;
+        }
+
+        const QByteArray importData = importFile.readAll();
```
Venemo: Perhaps add a file size check, so it won't be possible to accidentally try to import insanely huge files.

tools/secrets-tool/commandhelper.cpp
```diff
@@ -467,6 +468,54 @@ void CommandHelper::start(const QString &command, const QStringList &args)
...
+        if (importData.isEmpty()) {
+            qInfo() << "Unable to read data from import file!";
```
Venemo: Re-word the error message. It is also possible that it could read data from the file, but the file was simply empty.

tools/secrets-tool/commandhelper.cpp
chriadam: Done

tools/secrets-tool/commandhelper.cpp
chriadam: Done

chriadam:
Fixed review comments, rebased, merging.  Thanks!

PR #105, opened 2018-05-15
[sailfish-crypto] Rework Crypto unit tests. Contributes to JB#40058

blammit:
…nfigurable. Contributes to JB#40058

- Extract some common test code into CryptoTest
- Make the unit test plugins configurable via environment variables

### DIFF
--- a/tests/Crypto/cryptotest.cpp
+++ b/tests/Crypto/cryptotest.cpp
@@ -1,0 +1,290 @@
+/*
+ * Copyright (C) 2017 Jolla Ltd.
+ * Contact: Bea Lam <bea.lam@jollamobile.com>
+ * All rights reserved.
+ * BSD 3-Clause License, see LICENSE.
+ */
+
+#include "cryptotest.h"
+
+#include "Secrets/deletecollectionrequest.h"
+
+#include <QtDebug>
+
+using namespace Sailfish::Crypto;
+
+CryptoTest::CryptoTest(QObject *parent)
+    : QObject(parent)
+    , m_devRandom(nullptr)
+{
+}
+
+void CryptoTest::qtest_init()
+{
+
+}
+
+void CryptoTest::qtest_cleanup()
+{
+    while (m_populatedCollections.size() > 0) {
+        auto testCollection = m_populatedCollections.takeLast();
+        QDBusPendingReply<Result> reply = m_smp.deleteCollection(
+                    testCollection.name,
+                    testCollection.storagePlugin,
+                    testCollection.userInteractionMode);
+        WAIT_FOR_DBUS_REPLY(reply);
+    }
+}
+
+QByteArray CryptoTest::createRandomTestData(int size)
+{
+    if (!m_devRandom) {
+        m_devRandom = new QFile("/dev/urandom", this);
+        m_devRandom->open(QIODevice::ReadOnly);
+    }
+    if (m_devRandom->isOpen()) {
+        return m_devRandom->read(size);
+    }
+    qWarning() << "Cannot open" << m_devRandom->fileName();
+    return QByteArray();
+}
+
+QByteArray CryptoTest::generateInitializationVector(Sailfish::Crypto::CryptoManager::Algorithm algorithm,
+                                                    Sailfish::Crypto::CryptoManager::BlockMode blockMode)
+{
+    if (algorithm != CryptoManager::AlgorithmAes || blockMode == CryptoManager::BlockModeEcb) {
+        return QByteArray();
+    }
+    switch (blockMode) {
+        case CryptoManager::BlockModeGcm:
+            return createRandomTestData(12);
+        case CryptoManager::BlockModeCcm:
+            return createRandomTestData(7);
+    default:
+        break;
+    }
+    return createRandomTestData(16);
+}
+
+QByteArray CryptoTest::generateRsaPlaintext(Sailfish::Crypto::CryptoManager::EncryptionPadding padding, int keySize)
+{
+    if (padding == CryptoManager::EncryptionPaddingNone) {
+        // Otherwise OpenSSL will complain about too small / too large data size.
+        // See https://stackoverflow.com/questions/17746263/rsa-encryption-using-public-key-data-size-based-on-key
+        QByteArray plaintext = createRandomTestData(keySize / 8 - 1);
+        plaintext.prepend('\0');
+        return plaintext;
+    } else if (padding == CryptoManager::EncryptionPaddingRsaOaep) {
+        // Otherwise OpenSSL will complain about too small / too large data size.
+        return createRandomTestData(keySize / 32);
+    }
+
+    return createRandomTestData(42);
+}
+
+bool CryptoTest::allCharactersAreNull(const QString &s)
+{
+    for (auto c : s) {
+        if (c != '\0') {
+            return false;
+        }
+    }
+    return true;
+}
+
+Sailfish::Crypto::Key CryptoTest::createTestKey(int keySize,
+                                                Sailfish::Crypto::CryptoManager::Algorithm algorithm,
+                                                Sailfish::Crypto::Key::Origin origins,
+                                                Sailfish::Crypto::CryptoManager::Operations operations,
+                                                Key::Identifier keyIdentifier)
+{
+    Key key;
+    key.setSize(keySize);
+    key.setAlgorithm(algorithm);
+    key.setOrigin(origins);
+    key.setOperations(operations);
+    key.setFilterData(QLatin1String("test"), QLatin1String("true"));
+    key.setIdentifier(keyIdentifier);
+    return key;
+}
+
+void CryptoTest::addCryptoTestData(const TestPluginMap &plugins,
+                                   Key::Origin keyOrigin,
+                                   Sailfish::Crypto::CryptoManager::Operations operations,
+                                   Key::Identifier keyIdentifier,
+                                   const QByteArray &plaintext,
+                                   const TestRequests &testRequests)
+{
+    QTest::addColumn<TestPluginMap>("plugins");
+    QTest::addColumn<CryptoManager::BlockMode>("blockMode");
+    QTest::addColumn<CryptoManager::EncryptionPadding>("padding");
+    QTest::addColumn<Sailfish::Crypto::Key>("keyTemplate");
+
+    QTest::addColumn<QByteArray>("authData");
+    QTest::addColumn<QByteArray>("plaintext");
+    QTest::addColumn<QByteArray>("initVector");
+    QTest::addColumn<CryptoTest::TestRequests>("testRequests");
+
+    QByteArray authData16 = createRandomTestData(16);
+    QByteArray plaintextData = plaintext.isEmpty() ? "Test plaintext data" : plaintext;
+    QByteArray initVector;
+
+
+    // AES algorithm:
+
+    initVector = generateInitializationVector(CryptoManager::AlgorithmAes, CryptoManager::BlockModeEcb);
+    QTest::newRow("AES ECB 128-bit")
+            << plugins << CryptoManager::BlockModeEcb << CryptoManager::EncryptionPaddingNone
+            << createTestKey(128, CryptoManager::AlgorithmAes, keyOrigin, operations, keyIdentifier)
+            << QByteArray() << plaintextData << initVector << testRequests;
+    QTest::newRow("AES ECB 192-bit")
+            << plugins << CryptoManager::BlockModeEcb << CryptoManager::EncryptionPaddingNone
+            << createTestKey(192, CryptoManager::AlgorithmAes, keyOrigin, operations, keyIdentifier)
+            << QByteArray() << plaintextData << initVector << testRequests;
+    QTest::newRow("AES ECB 256-bit")
+            << plugins << CryptoManager::BlockModeEcb << CryptoManager::EncryptionPaddingNone
+            << createTestKey(256, CryptoManager::AlgorithmAes, keyOrigin, operations, keyIdentifier)
+            << QByteArray() << plaintextData << initVector << testRequests;
+
+    initVector = generateInitializationVector(CryptoManager::AlgorithmAes, CryptoManager::BlockModeCbc);
+    QTest::newRow("AES CBC 128-bit")
+            << plugins << CryptoManager::BlockModeCbc << CryptoManager::EncryptionPaddingNone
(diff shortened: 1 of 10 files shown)


tests/Crypto/cryptotest.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+bool CryptoTest::needsAuthentication(Sailfish::Crypto::CryptoManager::Algorithm algorithm,
+                                     Sailfish::Crypto::CryptoManager::BlockMode blockMode)
+{
+    if (m_testPluginCrypto == CryptoManager::DefaultCryptoPluginName + QLatin1String(".test")) {
+        return algorithm == Sailfish::Crypto::CryptoManager::AlgorithmAes
+                && (blockMode == Sailfish::Crypto::CryptoManager::BlockModeGcm
+                    || blockMode == Sailfish::Crypto::CryptoManager::BlockModeCcm);
+    }
+    qWarning() << "Test is not using the default test plugin, assuming authentication data is not required";
```
chriadam: Why do we only test the authentication data flows with the default plugin?

tests/Crypto/cryptotest.cpp
```diff
@@ -0,0 +1,158 @@
...
+    m_testPluginCryptoStorage = findPluginName("TEST_PLUGIN_CRYPTO_STORAGE", CryptoManager::DefaultCryptoStoragePluginName + QLatin1String(".test"));
+    m_testPluginStorage = findPluginName("TEST_PLUGIN_STORAGE", Sailfish::Secrets::SecretManager::DefaultStoragePluginName + QLatin1String(".test"));
+    m_testPluginEncryptedStorage = findPluginName("TEST_PLUGIN_ENCRYPTED_STORAGE", Sailfish::Secrets::SecretManager::DefaultEncryptedStoragePluginName + QLatin1String(".test"));
+    m_testPluginAuth = findPluginName("TEST_PLUGIN_AUTH", Sailfish::Secrets::SecretManager::DefaultAuthenticationPluginName + QLatin1String(".test"));
+    m_testPluginInAppAuth = findPluginName("TEST_PLUGIN_IN_APP_AUTH", Sailfish::Secrets::SecretManager::InAppAuthenticationPluginName + QLatin1String(".test"));
+    m_testPluginEncryption = findPluginName("TEST_PLUGIN_ENCRYPTION", Sailfish::Secrets::SecretManager::DefaultEncryptionPluginName + QLatin1String(".test"));
+    m_testPluginUsbToken = findPluginName("TEST_PLUGIN_USB_TOKEN", QStringLiteral("org.sailfishos.secrets.plugin.cryptostorage.exampleusbtoken.test"));
```
chriadam: While I like this idea in theory, in practice different plugins will have different levels of support for different operations and permutations of arguments - which means that specifying some different plugin via an environment variable is likely to result in a variety of unit test failures which shouldn't actually be considered failures (because different plugins support different semantics).  An alternative might be to pass the plugin names as QTest columns in the various _data() methods, along with expected results for different operations.  That way, other plugin implementors can add their own unit tests to enforce appropriate behaviour for their plugin simply by adding a data row to the appropriate _data() column.

tests/Crypto/cryptotest.cpp
blammit: I remember discussing that plugins would have their own authentication requirement configurations for custom algorithms and block modes, which we can't know. What do you think is a good way to handle that?

tests/Crypto/cryptotest.cpp
blammit: > specifying some different plugin via an environment variable is likely to result in a variety of 
> unit test failures which shouldn't actually be considered failures (because different plugins 
> support different semantics)

What sort of semantics? Does that mean the unit test itself may have to be changed per-plugin? How should it be reconfigured or skipped in those cases?

I did want to avoid adding _data() per-plugin as it means a lot of duplicated setup code, since QtTest doesn't support adding a set of test data on top of all current sets of test data (i.e. adding a set of providers across all the current test data configurations) but we can do that if necessary. 

If multiple plugins need to be supported across every unit test, then I suggest having an ini file that specifies the plugins for each plugin type (TEST_PLUGIN_AUTH etc).

tests/Crypto/cryptotest.cpp
chriadam: I think that it should be denoted by a QTest data column value.  The unit test itself should then do something like:

if (operationXSupported) {
    // ... test operation X ...
    QCOMPARE(resultValue, expectedOperationXResultValue);
}

tests/Crypto/cryptotest.cpp
chriadam: Yes, the unit tests need to test different things on a per-plugin basis, most likely.  That is, some plugins won't support some operations, or some block modes, or whatever, while others will.

So we cannot simply use an environment variable to specify the plugin name, in most cases, as each plugin may implement different things.  E.g. what if the OMP plugin doesn't support AES at all, but only supports GOST?  Or doesn't support SHA256 but supports SHA3_512 only?  etc.  Thus, the plugin name needs to be specified in the test data row, along with every input datum (e.g. algorithm, block mode, digest function, data to encrypt, etc) and every output datum (expected values, etc) .

What do you mean by "adding _data() per-plugin" there?  I just meant something like: each test should test some specific functionality.  So each data row for that test should include:

a) the name of the plugin to be tested
c) various inputs to specify the operation parameters
d) the expected result values for the given inputs

If a single unit test tests multiple independent features (e.g. the stored key requests which test generating stored keys, filtering stored keys, and using stored keys via encrypt/decrypt) it may be that a plugin supports some of those stages (e.g. generating and storing, and using the stored keys via encrypt/decrypt) but not others (e.g. it may not support filtering stored keys), and in that case the "does the plugin support this feature" flag needs to be passed as a row data also.  For example:

a) the name of the plugin to be tested
b) input data required to test generating stored keys
c) the expected result from that operation
d) whether the plugin supports filtering
e) input data required to test filtering
f) the expected result from filtering
g) whether the plugin supports encrypt operations
h) input data required to test encrypt operations
i) the expected result from encrypting
j) whether the plugin supports decrypt operations
k) input data required to test decrypt operations
l) the expected result from decrypting.

Then the test itself should look something like:

```
generateStoredKey(input params);
QCOMPARE(result, expectedGSKResult);

if (pluginSupportsFiltering) {
    filterKeys(someFilterParameters);
    QCOMPARE(results, expectedFilterResults);
}

if (pluginSupportsEncrypt) {
    encrypt(storedKey, inputData);
    QCOMPARE(results, expectedEncryptResults);

    if (pluginSupportsDecrypt) {
        decrypt(storedKey, encryptedData);
        QCOMPARE(decryptedData, originalEncryptInputData);
    }
}
```

tests/Crypto/cryptotest.cpp
blammit: Okay, yep the plugin test configuration could specify the algorithms and block modes that require authentication so that we can configure this accordingly.

tests/Crypto/cryptotest.cpp
chriadam: It previously existed, Joona asked me to remove it, as it complicated the API, and clients should know the semantics of the plugin they are attempting to interact with.  I actually agree with his reasoning about that: because the semantics of different plugins are different, merely reporting it programmatically is unlikely to be enough to allow a client to write code which has no knowledge of that plugin anyway (a good example is: what if the plugin requires a specific customParameter to be set?  do we also need some way to programmatically report that?)

Also, previously we reported it as a map from algorithm (e.g. AES) to QVector<BlockMode> or to QVector<SomeEncryptionPadding> etc - but this doesn't tell the complete picture, as a plugin may support AES with block mode CBC but only with some specific padding (like None), and also AES with a different block mode GCM, but only with some other specific padding (like PKCS#11 or whatever).  Basically, in order to report it effectively, programmatically, we would have some insanely complex lookup table (which then has to be defined by the plugin implementor!) in order to be useful...

blammit: Yes it sounds like it would be quite complex to have a plugin API that provides useful information about exactly what the plugin supports, so it would be too much trouble.

> etc - but this doesn't tell the complete picture, as a plugin may support AES with block mode 
> CBC but only with some specific padding (like None), and also AES with a different block mode 
> GCM, but only with some other specific padding (like PKCS#11 or whatever). 

If the configuration is purely just pluginSupportsEncrypt, pluginSupportsDecrypt etc. then we wouldn't be able to support these sorts of requirements in the tests. Is the plan to leave out this sort of detailed configuration in the tests for now?

chriadam: It's not needed, because that configuration is specified in the test itself.  That is, if you're writing some plugin and your plugin doesn't support AES CBC, then you either don't add a test data row for AES CBC, or alternatively you add a test data row for AES CBC which ensures that the return value is a failure with the expected error code.

blammit:
Still WIP, currently reworking following PR discussion.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -280,125 +264,131 @@ void tst_cryptorequests::getPluginInfo()
...
-    QCOMPARE(foundTestCryptoStoragePluginInC, true);
-    QCOMPARE(foundTestCryptoStoragePluginInS, true);
-    QCOMPARE(foundTestUsbPlugin, true);
+void tst_cryptorequests::randomData_data()
+{
+    QTest::addColumn<TestPluginMap>("plugins");
+    QTest::addColumn<QString>("csprngEngineName");
+    QTest::addColumn<int>("generateRandomDataSize");    // zero if plugin does not support GenerateRandomDataRequest
+    QTest::addColumn<double>("seedRandomDataEntropyEstimate");
+    QTest::addColumn<QByteArray>("seedData");   // empty if plugin does not support SeedRandomDataGeneratorRequest
```
chriadam: Probably also need QVariantMap customParameters column (for all _data() methods) to allow easily adding such custom parameters to requests where required by the backend (e.g. I know the gpg plugin has need for such custom parameters in some cases)

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -280,125 +264,131 @@ void tst_cryptorequests::getPluginInfo()
...
+
+    TestPluginMap plugins;
+    plugins.insert(CryptoTest::CryptoPlugin, DEFAULT_TEST_CRYPTO_PLUGIN_NAME);
+
+    QTest::newRow("DefaultCryptoPlugin")
+            << plugins << GenerateRandomDataRequest::DefaultCsprngEngineName
+            << 2048 << 0.5 << QByteArray("seed");
```
chriadam: Could also add row for the DEFAULT_TEST_CRYPTOSTORAGE_PLUGIN_NAME (which does not support seeding the rng), and other plugins etc.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -280,125 +264,131 @@ void tst_cryptorequests::getPluginInfo()
...
-    QCOMPARE(foundTestCryptoStoragePluginInS, true);
-    QCOMPARE(foundTestUsbPlugin, true);
+void tst_cryptorequests::randomData_data()
+{
+    QTest::addColumn<TestPluginMap>("plugins");
+    QTest::addColumn<QString>("csprngEngineName");
+    QTest::addColumn<int>("generateRandomDataSize");    // zero if plugin does not support GenerateRandomDataRequest
+    QTest::addColumn<double>("seedRandomDataEntropyEstimate");
+    QTest::addColumn<QByteArray>("seedData");   // empty if plugin does not support SeedRandomDataGeneratorRequest
+
```
chriadam: I guess we should also add a column like "expectedResult" which would be a Result(Result::Succeeded) if the plugin should support this operation, or Result(Result::OperationNotSupportedError, QString()) for plugins which don't support it.  The actual test should compare just the result.code() and result.errorCode() I guess.

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -280,125 +264,131 @@ void tst_cryptorequests::getPluginInfo()
...
-
-    // ensure that we get different random data to the original set
-    grdr.startRequest();
-    WAIT_FOR_FINISHED_WITHOUT_BLOCKING(grdr);
-    QByteArray seededData = grdr.generatedData();
-    QCOMPARE(seededData.size(), 2048);
-    QVERIFY(randomData != seededData);
+}
 
+void tst_cryptorequests::randomData_dev_urandom()
```
chriadam: This method can probably be removed, and should be replaced by simply adding a new row to the randomData_data() above which sets the engine to "/dev/urandom", I think

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -509,152 +498,163 @@ void tst_cryptorequests::generateKeyEncryptDecrypt()
...
+            << true;
 }
 
 void tst_cryptorequests::signVerify()
 {
-    QFETCH(CryptoManager::Algorithm, algorithm);
+    QFETCH(TestPluginMap, plugins);
+    QFETCH(Key, keyTemplate);
+    QFETCH(KeyPairGenerationParameters, keyPairGenParams);
     QFETCH(CryptoManager::DigestFunction, digestFunction);
```
chriadam: signaturePadding should also be passed as a column value here

tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
blammit: We could, though this test checks the generated random number, which isn't done in the randomData() test since that just accepts any non-null byte array, so we would lose that in the test.

blammit:
Made further extensive changes to allow specifying error results for all tests, and to add custom parameters for all requests.

Once this PR is merged, the Secrets/ tests should be adjusted similarly for consistency.

Venemo:
@blammit Can you please rebase this PR on top of latest master? Is this ready to get merged? If yes, I can review it.

blammit:
@Venemo I discussed this with @chriadam and as this is a significant change to the unit tests it may be better to leave it until after the major release is branched.

chriadam:
Oops I completely forgot about this one after the release was branched.  I have now rebased this one and fixed some conflicts / issues.  LGTM, merging now.

PR #106, opened 2018-05-16
[sailfish-crypto] Trigger collection unlock flow for crypto operations. Contributes to JB#40058

chriadam:
If a collection which contains a key is currently locked (due to the
unlock semantic being AccessRelock, for example), then if a client
attempts to use that key (e.g. to encrypt/decrypt/sign/verify) the
user should be prompted to unlock the collection in order to allow
the key to be retrieved in order to perform the operation.

This commit ensures that this occurs, where previously the operations
would have failed with a CollectionLocked error.

Contributes to JB#40058

### DIFF
--- a/daemon/CryptoImpl/cryptopluginfunctionwrappers.cpp
+++ b/daemon/CryptoImpl/cryptopluginfunctionwrappers.cpp
@@ -14,6 +14,41 @@ using namespace Sailfish::Crypto;
 using namespace Sailfish::Crypto::Daemon::ApiImpl;
 using namespace Sailfish::Secrets::Daemon::Util;
 
+namespace {
+    Sailfish::Secrets::Result unlockCollection(CryptoStoragePluginWrapper *w,
+                                               const QString &collectionName,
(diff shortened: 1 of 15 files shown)


chriadam:
The test for this is just the manual-test.sh script, as it creates keys in AccessRelock collections (thus when performing sign, verify, encrypt or decrypt steps with stored (or imported) keys, it should prompt the user to unlock the collection prior to performing those operations).

chriadam:
WIP, still fixing unit tests etc

chriadam:
This is now ready for review.

rainemak:
LGTM

rainemak:
I'd like to get @Venemo 's review as well.

chriadam:
It is a public holiday in Hungary today, so I don't think Venemo will be working.  We really cannot afford to delay this PR any longer... It is blocking j-doge's testing, and we need as much testing and validation as possible before the release...

chriadam:
Merging now, as per discussion on IRC.

PR #108, opened 2018-05-17
[sailfish-secrets] Make PluginInfo read-only, fixes #90.

### LINKED ISSUE #90: Proposition to make PluginInfo API reflect its read-only purpose
The current API of {Crypto,Secret}/PluginInfo provides setters for read-only data such as plugin name or version. A grep in the code shows that ```setVersion()``` for instance is used in:

<pre>
./daemon/controller.cpp:        info.setVersion(plugin->version());
./lib/Crypto/serialization.cpp:    info.setVersion(version);
./lib/Secrets/serialization.cpp:    info.setVersion(version);
</pre>

which are purely internal methods. Maybe it could be clearer to remove these setters and pass these values at construct time.

Additional side note about PluginInfo API: I guess it will be used in UI to display selectors for crypto plugin to use when performing a signature action for instance. In that case, is it not too restrictive to make it a Q_GADGET? The available property is clearly something that may change and should notify on change. I know that plugins currently don't have the possibility to notify the main thread of changes in general, but they will eventually have and will it be too late to change from Q_GADGET to Q_OBJECT, or not?

dcaliste:
This PR removes the setters method from PluginInfo structure to better reflect its read-only purpose, as discussed in #90.

### DIFF
--- a/daemon/controller.cpp
+++ b/daemon/controller.cpp
@@ -146,13 +146,8 @@ Sailfish::Secrets::Daemon::Controller::pluginInfoForPlugins(
 {
     QMap<QString, Sailfish::Secrets::PluginInfo> infos;
     for (Sailfish::Secrets::PluginBase *plugin : plugins) {
-        Sailfish::Secrets::PluginInfo info;
-
         // metadata reporting occurs in main thread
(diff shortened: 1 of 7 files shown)


chriadam:
LGTM tyvm

PR #111, opened 2018-05-21
[sailfish-secrets] Add support for keep-unlocked semantics to secrets-tool. Contributes to JB#36797

chriadam:
This commit adds support for specifying the lock semantic which should
be applied to the collection or standalone secret being created
with the secrets-tool.

It also adds a new manual test script which performs operations on
a "keep-unlocked" collection, to test that the semantic is applied
appropriately (and so the user is not prompted to enter the unlock
code for every operation, as they are for manual-test.sh).

Finally, it supports specifying secret data via the tool rather
than requiring that data to be entered via UI prompt.

Contributes to JB#36797


tools/secrets-tool/commandhelper.cpp
```diff
@@ -189,49 +189,71 @@ void CommandHelper::start(const QString &command, const QStringList &args)
...
     } else if (command == QStringLiteral("--create-collection")) {
-        Sailfish::Secrets::CreateCollectionRequest *r = new Sailfish::Secrets::CreateCollectionRequest;
         bool devicelock = false;
-        if (args.size() == 4
-                || (args.size() == 3 && args.value(0) == QStringLiteral("--devicelock"))) {
-            if (args.value(0) != QStringLiteral("--devicelock")) {
-                qInfo() << "Invalid lock mode specified!";
+        bool keepUnlocked = false;
+        QStringList ccArgs(args);
+        if (ccArgs.first() == QStringLiteral("--devicelock")) {
```
rainemak: There should be !isEmpty() checks before calling first().

rainemak: Thanks for clarifying just checked from github's interface

PR #112, opened 2018-05-21
[sailfish-secrets] Add InteractionViewUserCanceledError error code. Contributes to JB#36797

chriadam:
If the user cancels a user input dialog, we should report this as
differently to other failures (e.g. dbus communication failure).
This commit adds a specific error code to denote this case, and
ensures that the passwordagent plugin returns that error code
from user input flows which are canceled.

Contributes to JB#36797


rainemak:
LGTM

chriadam:
Updated to support the polkit.dismissed case and also added a note in the InteractionRequest documentation about the InteractionViewUserCanceledError error code.  Rebased.  Will merge, thanks.

PR #114, opened 2018-05-22
[sailfish-secrets] Report collection lock state from CollectionNamesRequest. Contributes to JB#36797

chriadam:
This commit modifies the CollectionNamesRequest so that the lock
state of the collection is exposed, allowing the client to
modify its behaviour depending on the lock state of the collection,
rather than reacting to CollectionLocked errors.

Contributes to JB#36797

### DIFF
--- a/daemon/SecretsImpl/pluginwrapper.cpp
+++ b/daemon/SecretsImpl/pluginwrapper.cpp
@@ -172,7 +172,9 @@ bool StoragePluginWrapper::initialize(const QByteArray &masterLockKey)
     }
 
     QStringList cnames;
-    Result result = collectionNames(&cnames);
+    QVariantMap cnamesMap;
+    Result result = collectionNames(&cnamesMap);
+    cnames = cnamesMap.keys();
     if (result.code() != Result::Succeeded) {
         return false;
     }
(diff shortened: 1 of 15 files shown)


chriadam:
Ready for review.  The fact that the dbus return value type is QVariantMap is ugly, but functional... this ugliness is hidden from the client API, so is probably ok.

daemon/SecretsImpl/pluginwrapper.cpp
```diff
@@ -561,9 +571,22 @@ EncryptionPlugin::EncryptionAlgorithm EncryptedStoragePluginWrapper::encryptionA
...
 {
-    return m_encryptedStoragePlugin->collectionNames(names);
+    QStringList cnames;
+    Result result = m_encryptedStoragePlugin->collectionNames(&cnames);
+    if (result.code() == Result::Succeeded) {
+        for (const QString &cname : cnames) {
+            bool locked = false;
+            Result lockedResult = m_encryptedStoragePlugin->isCollectionLocked(cname, &locked);
+            if (lockedResult.code() != Result::Succeeded) {
+                // assume locked, otherwise ignore the error.
```
Venemo: Is it wise to assume that any error here means that it's locked? I would suggest to check the error code and if there is an error here, return that error code to the caller.

chriadam: It is less useful to fail the entire operation if an attempt to read the lock state of a single collection fails, than to return at least the collection names (which have, at this point, already been retrieved without any error).  Assuming that the collection is locked is the "safe" option then.

chriadam:
The name of the request is correct as per the API it offers (with collectionNames property).  Another option would be to encapsulate the name+lockState for each collection into a CollectionInfo struct, and change the name to CollectionInfoRequest, however that is an invasive change at this point considering that everything needs to be merged and tagged tonight.  Personally, I prefer leaving the API as-is (to the client, the entire API change was adding a single Q_INVOKABLE method).

rainemak:
LGTM

daemon/SecretsImpl/pluginwrapper.cpp
Venemo: In this case I suggest 3 states: locked ok, unlocked ok, errored --- the 3rd state telling the client that there was an error trying to determine the state of this collection, thus it probably can't be used.

PR #123, opened 2018-05-25
[sailfish-secrets] Initialize passwordagent once accepted. Contributes to JB#36797

chriadam:
We build both a "normal" and a "test" version of various plugins.
Upon daemon startup, we load all plugins, and then unload any which
don't match the current mode (autotest = true/false).

This has the unfortunate side effect that the test version of the
passwordagent plugin will overwrite the socket file originally
created by the normal version of the password agent plugin, before
being unloaded.

This commit ensures that we allow plugins to perform initialization
in a specific initialize() method which is only called once the
plugin has been "accepted" for use by the daemon according to the
current mode.

Contributes to JB#36797

### DIFF
--- a/lib/Secrets/Plugins/extensionplugins.h
+++ b/lib/Secrets/Plugins/extensionplugins.h
@@ -35,6 +35,8 @@ public:
     PluginBase();
     virtual ~PluginBase();
 
+    virtual void initialize();
+
     virtual QString displayName() const = 0;
     virtual QString name() const = 0;
     virtual int version() const = 0;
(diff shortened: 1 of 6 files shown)


lib/Secrets/Plugins/extensionplugins.h
```diff
@@ -35,6 +35,8 @@ class SAILFISH_SECRETS_API PluginBase
     PluginBase();
     virtual ~PluginBase();
 
+    virtual void initialize();
```
adenexter: I thought this could return a bool indicating success/failure, and a failed plugin could be removed. But I guess it makes a sort of sense to keep them around to return meaningful errors if someone tries to utilize the plugin.

adenexter:
LGTM

Issue #122, opened 2018-05-24
Add possibility to enter custom parameters with the secrets-tool CLI


dcaliste:
Would be nice to be able to type something like:
<pre>secrets-tool --sign openpgp "Joe Dalton" AED123 -o Text-Mode=true -o With-Armor=false</pre>

This issue, is more a reminder for me about it to propose a PR later.

dcaliste:
Implemented in #124.

PR #124, opened 2018-05-29
[sailfish-secrets] Add support to pass custom parameters on CLI, fixes Github#122.

dcaliste:
Add support to add ```-o key=value``` options in the command line for crypto requests. This is then passed to the crypto API as customParameters. It implements #122.

### DIFF
--- a/tools/secrets-tool/commandhelper.cpp
+++ b/tools/secrets-tool/commandhelper.cpp
@@ -65,6 +65,39 @@ static Sailfish::Crypto::CryptoManager::DigestFunction digestEnum(const QString 
     return Sailfish::Crypto::CryptoManager::DigestSha512;
 }
 
+static QVariantMap toCustomParameters(const QStringList &options)
+{
+    QVariantMap customs;
+    for (QList<QString>::ConstIterator it = options.constBegin(); it != options.constEnd(); it++) {
+        int iEqual = it->indexOf('=');
+        if (iEqual > 0 && iEqual < it->length() - 1) {
+            const QString key = it->left(iEqual);
+            const QString value = it->mid(iEqual + 1);
+            bool ok;
+            QVariant vd(value.toDouble(&ok));
+            if (ok) {
+                customs.insert(key, vd);
+            } else {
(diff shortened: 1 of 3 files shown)


tools/secrets-tool/commandhelper.cpp
```diff
@@ -65,6 +65,39 @@ static Sailfish::Crypto::CryptoManager::DigestFunction digestEnum(const QString
...
+            bool ok;
+            QVariant vd(value.toDouble(&ok));
+            if (ok) {
+                customs.insert(key, vd);
+            } else {
+                QVariant vi(value.toInt(&ok));
+                if (ok) {
+                    customs.insert(key, vi);
+                } else {
+                    if (value.compare("true", Qt::CaseInsensitive) == 0)
```
chriadam: Sailfish OS coding style is to use brackets even around single-line conditional blocks.

tools/secrets-tool/commandhelper.cpp
```diff
@@ -65,6 +65,39 @@ static Sailfish::Crypto::CryptoManager::DigestFunction digestEnum(const QString
     return Sailfish::Crypto::CryptoManager::DigestSha512;
 }
 
+static QVariantMap toCustomParameters(const QStringList &options)
+{
+    QVariantMap customs;
+    for (QList<QString>::ConstIterator it = options.constBegin(); it != options.constEnd(); it++) {
+        QStringList keyValue(it->split('='));
```
chriadam: instead of using split, could instead use split+join, to allow options like '-o someKey="someValueWith=Sign"' (that is, just split out the LHS of the first = as the key, and join the rest to form the value.

chriadam:
A couple of minor comments, aside from those LGTM - thanks!

tools/secrets-tool/commandhelper.cpp
dcaliste: Ah, ok, I was using the same convention than for sailfish-office (which is following the one of Qt), advicing not to put brackets if all conditions branches fit on one line.

I've updated the PR.

PR #126, opened 2018-06-05
[sailfish-secrets] Improve application id determination. Contributes to JB#42038

chriadam:
Attempt to read application identifier from cgroup, then from exe link,
then fall back to cmdline parsing.

Contributes to JB#42038


chriadam:
WIP, readlink will fail unless the user+group of the calling application is the same as the daemon, presumably because sysctl -w fs.protected_symlinks=1 is set for the procfs...

daemon/SecretsImpl/applicationpermissions.cpp
```diff
@@ -12,46 +12,126 @@
...
         qCDebug(lcSailfishSecretsDaemon) << "zero pid, assuming privileged!";
         return platformApplicationId();
     }
 
-    const QString pidFile = QString::fromLatin1("/proc/%1/cmdline").arg(pid);
-    if (!QFile::exists(pidFile)) {
-        qCWarning(lcSailfishSecretsDaemon) << "no such pid:" << pid;
-        return QString();
+    const QString cgroupName = readBoosterCgroup(pid);
+    if (!cgroupName.isEmpty()) {
```
aleksey-mikhaylov: Blank cgroup is "/" not empty string afaik.

aleksey-mikhaylov:
It should not be forgotten that booster cgroup should be mount at startup by separate unit file or by other means. Now there is no such mechanism.

daemon/SecretsImpl/applicationpermissions.cpp
chriadam: Updated check to also ignore if equal to QStringLiteral("/")

chriadam:
Updated and tested, so it should work now (for cases where the name=booster cgroup is present...)

Regarding your comment above about mounting booster cgroup at startup by separate unit file, where does that change need to be made?

daemon/SecretsImpl/applicationpermissions.cpp
```diff
@@ -12,46 +12,125 @@
...
+                }
+            }
+        }
 
-    // special case handling of the pid 999999 for demonstration testing purposes only!
-    // TODO: remove this special case handling!
-    if (pid == 999999) {
-        qWarning() << "WARNING! Special case pid handling needs to be removed once access control / application management API exists!";
-        return QLatin1String("test-third-party-application");
+        return QString();
```
Venemo: Add a `qCDebug` here to indicate that the relevant cgroup was not found.

daemon/SecretsImpl/applicationpermissions.cpp
```diff
@@ -12,46 +12,125 @@
...
     }
 
+    QString readExeSymlink(pid_t pid)
+    {
+        struct stat sb;
+        const QString pidFile(QStringLiteral("/proc/%1/exe").arg(pid));
+        if (lstat(pidFile.toStdString().c_str(), &sb) == -1) {
```
Venemo: Maybe use `pidFile.toUtf8().data()` that way it's clearer what is going on.

daemon/SecretsImpl/applicationpermissions.cpp
```diff
@@ -12,46 +12,125 @@
...
+        }
+
+        QByteArray contents;
+        char data;
+        while (file.read(&data, 1)) {
+            if (data == '\0') {
+                data = ' ';
+            }
+            contents.append(data);
+        }
```
Venemo: Couldn't this be solved using `readAll` and then `replace()`?

daemon/SecretsImpl/applicationpermissions.cpp
```diff
@@ -12,46 +12,125 @@
...
+        const QString retn(QString::fromUtf8(contents).trimmed());
+        qCDebug(lcSailfishSecretsDaemon) << "caller with pid" << pid << "has cmdline applicationId:" << retn;
+        return retn;
+    }
+}
+
+QString Sailfish::Secrets::Daemon::ApiImpl::ApplicationPermissions::applicationId(pid_t pid) const
+{
     if (pid == 0) {
         qCDebug(lcSailfishSecretsDaemon) << "zero pid, assuming privileged!";
```
Venemo: Is this still a correct assumption here?

daemon/SecretsImpl/applicationpermissions.cpp
chriadam: done

daemon/SecretsImpl/applicationpermissions.cpp
chriadam: done

daemon/SecretsImpl/applicationpermissions.cpp
chriadam: Perhaps, but I don't want to make functional changes to that parsing code in this change

daemon/SecretsImpl/applicationpermissions.cpp
chriadam: Yes, this should not be any different now.

PR #127, opened 2018-06-07
[sailfish-secrets] Add new GOST DigestFunction enum values.

vityafx:
`DigestFunction` enum lacks GOST algorithms family in a way it could be used. This adds needed values.

Signed-off-by: Victor Polevoy <v.polevoy@omprussia.ru>


vityafx:
I hope I did not forget a place which I had to change.

chriadam:
Thanks!  Is something similar also needed for Algorithm (currently value 270 is just AlgorithmGost)?

chriadam:
Also EllipticCurve are their GOST curves defined?

Denis-Semakin:
@chriadam @vityafx 
> Thanks! Is something similar also needed for Algorithm (currently value 270 is just AlgorithmGost)?

I think it could be. E.g. there are two standards for encrypt()/decrypt(): Magma (old but still using) and Kuznechick (grasshopper in Engilsh).

> Also EllipticCurve are there GOST curves which should be defined?

I don't think that we have any "special" elliptic curves which should be marked with different name.

vityafx:
@Denis-Semakin @chriadam 

So I add `Magma` and `Grasshopper` into Algorithm?

Denis-Semakin:
@vityafx Yes

vityafx:
Added algorithms.

Denis-Semakin:
+1

vityafx:
Removed arguments after internal discussion that it probably not needed.

vityafx:
Nothing more is being discussed, you can merge. :)

PR #131, opened 2018-06-13
Add ability to change Qt logging categories

vityafx:
This commit adds "EnvironmentFile" option for the sailfish-secretsd unit file.
After these changes the service will be able to change logging categories at
runtime. The easiest way to do this is to install a special rpm logging package
which installs a file into this directory.

### DIFF
--- a/daemon/sailfish-secretsd.service
+++ b/daemon/sailfish-secretsd.service
@@ -7,6 +7,7 @@ Conflicts=shutdown.target jolla-actdead-charging.service
 [Service]
 Type=dbus
 BusName=org.sailfishos.secrets.daemon.discovery
+EnvironmentFile=-/var/lib/environment/sailfish-secretsd/*.conf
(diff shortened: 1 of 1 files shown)


Denis-Semakin:
LGTM

daemon/sailfish-secretsd.service
```diff
@@ -7,6 +7,7 @@ Conflicts=shutdown.target jolla-actdead-charging.service
 [Service]
 Type=dbus
 BusName=org.sailfishos.secrets.daemon.discovery
+EnvironmentFile=-/var/lib/environment/sailfish-secretsd/*.conf
```
chriadam: is the dash ("-" after the "=") intentional?

vityafx:
Yes.

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

Victor Polevoy

PR #132, opened 2018-06-14
[sailfish-secrets] Fix master lock key initialization. Contributes to JB#42034

chriadam:


chriadam:
This still currently requires that a crypto plugin is present which can perform key derivation (in order to implement the requirement from IZh regarding using PBKDF to generate a key from the input data, see his response to my email "Open question about the semantics of device-locked collections").

Perhaps we should allow the initialization to succeed even if no crypto plugin capable of performing key derivation is present, and simply store a dummy/null key in that case?

daemon/SecretsImpl/secrets.cpp
```diff
@@ -499,69 +504,149 @@ bool Daemon::ApiImpl::SecretsRequestQueue::generateKeyData(
...
     }
 
-    Sailfish::Crypto::Result bkdbKeyResult = cplugin->generateKey(
-                keyTemplate,
-                Sailfish::Crypto::KeyPairGenerationParameters(),
-                kdfParams,
-                QVariantMap(),
-                &bookkeepingdbKey);
+    // attempt to use the plugin specified in the environment variable
+    const QString envCPluginName = QString::fromLocal8Bit(qgetenv("SAILFISH_SECRETSD_MASTERLOCK_CRYPTOPLUGIN"));
```
Venemo: `SAILFISH_SECRETSD_MASTERLOCK_CRYPTOPLUGIN` should not be hardcoded here. I suggest to `#define` it in some header file. Also some documentation wouldn't hurt.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -702,14 +828,15 @@ bool Daemon::ApiImpl::SecretsRequestQueue::compareTestCipherText(
...
-            s = dataProtector.putData(testCipherText);
+            s = dataProtector.putData(cipherPluginName.toUtf8() + '\n' + testCipherText);
             return s == DataProtector::Success;
         } else {
             qCWarning(lcSailfishSecretsDaemon) << "Unable to read ciphertext data from nonexistent lock code check file";
             return false;
         }
     } else {
-        if (previousData != testCipherText) {
+        const QByteArray previousDataExceptPluginName = previousData.mid(previousData.indexOf('\n')+1);
```
Venemo: Bad tabulation of `)+1`. Should be `) + 1`.

Venemo:
> Perhaps we should allow the initialization to succeed even if no crypto plugin capable of performing key derivation is present, and simply store a dummy/null key in that case?

As far I see the default OpenSSL plugin should be always present. I would vote not to support installations where no crypto plugins are present.

daemon/SecretsImpl/secrets.cpp
chriadam: good point, done

daemon/SecretsImpl/secrets.cpp
chriadam: done

chriadam:
I agree that making the default plugin (for each plugin type) configurable is a good solution.

chriadam:
Thanks, rebased, merging now.

PR #133, opened 2018-06-27
[sailfish-secrets] Perform plugin name mapping. Contributes to JB#42192

chriadam:
This commit changes the client-side default plugin name variables
to be generic names (e.g. "plugin.storage.default") which then get
mapped to "real" plugin names by the daemon.

This allows the default plugins to be configurable (via .conf file
which provides environment variable values), allowing vendors to
supply their own default plugins which clients can use without
needing to modify client code.

Finally, unit tests are fixed to work with the changes.

Contributes to JB#42192

### DIFF
--- a/tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
+++ b/tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
@@ -287,7 +287,7 @@ void tst_cryptorequests::getPluginInfo()
     bool foundTestCryptoStoragePluginInS = false;
     bool foundTestUsbPlugin = false;
     for (auto p : r.cryptoPlugins()) {
-        if (p.name() == DEFAULT_TEST_CRYPTO_PLUGIN_NAME) {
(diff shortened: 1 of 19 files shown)


tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp
```diff
@@ -3361,17 +3361,19 @@ void tst_cryptorequests::importKey_data()
...
-            << promptToCancel
-            << Result::Failed
-            << Result::CryptoPluginKeyImportError
-            << QByteArray()
-            << QByteArray()
-            << 0
-            << Key::OriginUnknown
-            << CryptoManager::AlgorithmUnknown;
+
+    // this test should be enabled for manual testing only, user has to cancel the dialog.
```
Venemo: Is this added intentionally here? Looks like it's not related to the current topic.

chriadam: Intentional, I noticed the discrepancy when I updated this test to use the proper plugin.  Figured it is better to ensure that all tests pass as a result of the commit, rather than splitting the plugin mapping and unit test fixing over two separate commits.

PR #135, opened 2018-07-03
Add ImportStoredKeyRequest in QML export.

dcaliste:
The importation request was missing from QML. This PR adds it. It also:

* correct the Request::Status type when exported to QML, to avoid error like "unsupported type Request::Status for property status of Sailfish::Crypto::ImportStoredKeyRequest";
* log reason for metadata database initialisation failures;
* add a ```constructKeyTemplate()``` method to be able to write things like that in QML:
<pre>
ImportStoredKeyRequest {
    keyTemplate: cryptoManager.constructKeyTemplate(email.signingKeys[0],
                                                   "import",
                                                   "org.sailfishos.crypto.plugin.gnupg.openpgp")
}
</pre>

### DIFF
--- a/daemon/SecretsImpl/pluginwrapper.cpp
+++ b/daemon/SecretsImpl/pluginwrapper.cpp
@@ -172,12 +172,16 @@ bool StoragePluginWrapper::initialize(const QByteArray &masterLockKey)
     // step one: open the metadata database.
     if (!m_metadataDb.isOpen()) {
         if (!m_metadataDb.openDatabase(masterLockKey)) {
+            qCWarning(lcSailfishSecretsDaemon) << "cannot open metadata database:"
+                                               << m_metadataDb.errorMessage();
             return false;
         }
     }
 
     // step two: read current data from the plugin.
     if (m_storagePlugin->isLocked()) {
+        qCWarning(lcSailfishSecretsDaemon)
+            << "cannot read data from plugin: plugin is locked.";
         return false;
     }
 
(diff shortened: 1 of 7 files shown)


qml/Crypto/main.cpp
```diff
@@ -74,6 +76,11 @@ Sailfish::Crypto::Key Sailfish::Crypto::Plugin::CryptoManager::constructKey() co
     return Sailfish::Crypto::Key();
 }
 
+Sailfish::Crypto::Key Sailfish::Crypto::Plugin::CryptoManager::constructKeyTemplate(const QString &name, const QString &collectionName, const QString &storagePluginName) const
```
Venemo: This is essentially an overload of `constructKey` ― could you rename it as such?

daemon/SecretsImpl/pluginwrapper.cpp
```diff
@@ -545,6 +564,9 @@ bool EncryptedStoragePluginWrapper::initialize(const QByteArray &masterLockKey)
                 QStringList snames;
                 result = secretNames(cname, &snames);
                 if (result.code() != Result::Succeeded) {
+                    qCWarning(lcSailfishSecretsDaemon) << "cannot get secret names for"
+                                                       << cname << ":"
+                                                       << result.errorMessage();
```
Venemo: Please also print the result code in the warning.

dcaliste: I've added code wherever it apply. Thanks for the suggestion.

qml/Crypto/main.cpp
dcaliste: That's reasonable, done.

PR #136, opened 2018-07-04
[sailfish-secrets] Fix crash in device lock authentication flow. Contributes to JB#42295

chriadam:


plugins/passwordagentauthplugin/passwordagentplugin.cpp
```diff
@@ -523,9 +526,15 @@ Result PasswordAgentPlugin::beginAuthentication(
...
                 result = Result(Result::InteractionViewUserCanceledError,
                                 QLatin1String("The user canceled the authentication dialog"));
-            } else {
+            } else if (!QDBusReply<PolkitAuthorizationResult>(*watcher).value().details.keys().isEmpty()) {
                 result = Result(Result::IncorrectAuthenticationCodeError,
                                 QStringLiteral("Password Agent was unable to verify the authenticity of the user"));
+            } else {
+                // The calling application was not started by the lipstick user session
+                // so polkit cannot work for it (as the lipstick-security-ui is for that session).
+                // Instead, use devicelock API directly.
```
Venemo: Besides the comment,  I'd suggest to add a `qcDebug` message here. That would help if this comes up during debugging.

plugins/passwordagentauthplugin/passwordagentplugin.cpp
```diff
@@ -535,6 +544,49 @@ Result PasswordAgentPlugin::beginAuthentication(
...
+        uint callerPid,
+        qint64 requestId,
+        const Sailfish::Secrets::InteractionParameters::PromptText &promptText)
+{
+    QTimer *timeout = new QTimer(this);
+    timeout->setInterval(120000); // 2 minutes
+    timeout->setSingleShot(true);
+    connect(timeout, &QTimer::timeout, timeout, [this, timeout, callerPid, requestId] {
+        authenticationCompleted(callerPid, requestId,
+                                Result(Result::IncorrectAuthenticationCodeError,
```
Venemo: Is `IncorrectAuthenticationCodeError` the proper error code here? Sounds like this is basically a timeout error, and the error code and message should reflect that.

rpm/sailfish-secrets.spec
```diff
@@ -45,10 +45,12 @@ Summary:    Unit tests for the Sailfish OS Secrets Library
...
 BuildRequires:  pkgconfig(Qt5QuickTest)
 BuildRequires:  pkgconfig(libcrypto)
-Requires: qt5-qtdeclarative-import-qttest
-Requires: qt5-qtdeclarative-devel-tools
-Requires: openssl
+BuildRequires:  pkgconfig(nemodevicelock)
+Requires:   qt5-qtdeclarative-import-qttest
+Requires:   qt5-qtdeclarative-devel-tools
 Requires:   libsailfishsecrets = %{version}-%{release}
+Requires:   openssl
```
Venemo: Is openssl used by the auth plugin?

chriadam: it's used by the test openssl plugin, which is also part of this test package.

plugins/passwordagentauthplugin/passwordagentplugin.cpp
chriadam: No such result code currently exists, but you're right - I will add one.

plugins/passwordagentauthplugin/passwordagentplugin.cpp
chriadam: Good point, I will add this one.

dcaliste:
@chriadam : it works great now ! Many thanks.

Tested workflow:

* start secret daemon.
* set ```/desktop/sailfish/secrets/storeGnuPGPassphrases``` to true
* execute ```gpg2 -s toto```, system dialog appears and ask for secret passphrase.
* execute signing command again, operation is done without user interaction.
* restart secret daemon.
* execute signing command again, a system dialog appears to ask for unlock of secret.
* execute signing command again, operation is done without user interaction.

PR #137, opened 2018-07-05
Better handling of data corruption

Venemo:
This PR is a follow-up to the data protection PR which added the `DataProtector` class to the secrets daemon. This one deals with what happens when data corruption happens.

* Separate the cases when there is data corruption from when the data is otherwise inaccessible.
* When the data is irretrieveable, delete the data.
* When the data is irretrieveable, send a notification to the user.

Further idea:

* When just the lock code data is gone, maybe prompt the user to set up a new lock code. This is nice to have, but considering how unlikely it is that this actually happens, I have not implemented it yet. Our time is better spent elsewhere.

### DIFF
--- a/daemon/SecretsImpl/secrets.cpp
+++ b/daemon/SecretsImpl/secrets.cpp
@@ -38,6 +38,11 @@
 #define MAP_PLUGIN_NAMES(variable) ::mapPluginNames(m_requestQueue->controller(), variable)
 
 namespace {
+
+    const QString systemDataDirPath(QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + QStringLiteral("/system"));
+    const QString privilegedDataDirPath(systemDataDirPath + QStringLiteral("/privileged"));
+    const QString secretsDirPath(privilegedDataDirPath + QStringLiteral("/Secrets"));
+
     void specifyDummyMasterlockKeys(
             const QByteArray &lockCode,
             QByteArray *testCipherText,
@@ -818,9 +823,7 @@ bool Daemon::ApiImpl::SecretsRequestQueue::writeTestCipherText(
     // so this test should be safe.
     // TODO: check with crypto expert!
     // TODO: Should I just store a hash of the key instead?
-    const QString systemDataDirPath(QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/system/");
-    const QString privilegedDataDirPath(systemDataDirPath + QLatin1String("privileged") + "/");
-    const QString secretsDirPath(privilegedDataDirPath + QLatin1String("Secrets"));
+
     QDir secretsDir(secretsDirPath);
     if (!secretsDir.mkpath(secretsDirPath)) {
         qCWarning(lcSailfishSecretsDaemon) << "Permissions error: unable to create secrets directory:" << secretsDirPath;
@@ -845,9 +848,6 @@ bool Daemon::ApiImpl::SecretsRequestQueue::writeTestCipherText(
 bool Daemon::ApiImpl::SecretsRequestQueue::determineTestCipherPlugin(
         QString *cipherPluginName) const
 {
-    const QString systemDataDirPath(QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/system/");
-    const QString privilegedDataDirPath(systemDataDirPath + QLatin1String("privileged") + "/");
-    const QString secretsDirPath(privilegedDataDirPath + QLatin1String("Secrets"));
     QDir secretsDir(secretsDirPath);
     if (!secretsDir.mkpath(secretsDirPath)) {
         qCWarning(lcSailfishSecretsDaemon) << "Permissions error: unable to create secrets directory:" << secretsDirPath;
@@ -886,9 +886,7 @@ bool Daemon::ApiImpl::SecretsRequestQueue::compareTestCipherText(
     // so this test should be safe.
     // TODO: check with crypto expert!
     // TODO: Should I just store a hash of the key instead?
-    const QString systemDataDirPath(QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/system/");
-    const QString privilegedDataDirPath(systemDataDirPath + QLatin1String("privileged") + "/");
-    const QString secretsDirPath(privilegedDataDirPath + QLatin1String("Secrets"));
+
     QDir secretsDir(secretsDirPath);
     if (!secretsDir.mkpath(secretsDirPath)) {
         qCWarning(lcSailfishSecretsDaemon) << "Permissions error: unable to create secrets directory:" << secretsDirPath;
@@ -963,9 +961,6 @@ QByteArray Daemon::ApiImpl::SecretsRequestQueue::saltData() const
         return m_saltData;
     }
 
-    const QString systemDataDirPath(QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/system/");
-    const QString privilegedDataDirPath(systemDataDirPath + QLatin1String("privileged") + "/");
-    const QString secretsDirPath(privilegedDataDirPath + QLatin1String("Secrets"));
     QDir secretsDir(secretsDirPath);
     if (!secretsDir.mkpath(secretsDirPath)) {
         qCWarning(lcSailfishSecretsDaemon) << "Permissions error: unable to create secrets directory:" << secretsDirPath;
@@ -2599,11 +2594,6 @@ void Daemon::ApiImpl::SecretsRequestQueue::dealWithDataCorruption() const
     // NOTE: Right now we just delete all corrupted data.
     //       In the future if only the masterlock is broken we could just ask
     //       the user to set a new master lock (without clearing all data).
-
-    // Find secrets directory
-    const static QString systemDataDirPath(QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/system/");
-    const static QString privilegedDataDirPath(systemDataDirPath + QLatin1String("privileged") + "/");
-    const static QString secretsDirPath(privilegedDataDirPath + QLatin1String("Secrets"));
 
     // Remove entire secrets directory
     QDir secretsDir(secretsDirPath);
--- a/daemon/SecretsImpl/secrets.cpp
+++ b/daemon/SecretsImpl/secrets.cpp
@@ -836,7 +836,7 @@ bool Daemon::ApiImpl::SecretsRequestQueue::writeTestCipherText(
     DataProtector::Status s = dataProtector.putData(cipherPluginName.toUtf8() + '\n' + testCipherText);
     bool ok = (s == DataProtector::Success);
     if (!ok) {
-        qCWarning(lcSailfishSecretsDaemon) << "Can't write lock code data. DataProtector returned:" << s;
+        qCWarning(lcSailfishSecretsDaemon) << "writeTestCipherText: Can't write lock code data. DataProtector returned:" << s;
     }
 
     return ok;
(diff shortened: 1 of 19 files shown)


daemon/SecretsImpl/secrets.cpp
```diff
@@ -2586,3 +2594,23 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
     }
 }
 
+void Daemon::ApiImpl::SecretsRequestQueue::dealWithData() const
```
dcaliste: Didn't try to compile, but why the method is called dealWithData() here, while everywhere you call it with dealWithDataCorruption() (which sounds a better name).

daemon/SecretsImpl/secrets.cpp
```diff
@@ -2586,3 +2594,23 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
...
+{
+    // Find secrets directory
+    const static QString systemDataDirPath(QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation) + "/system/");
```
dcaliste: I think this directory is built somewhere else in the code (like the place where to create the databases). It would be better to have a method that return this path, so in case of a change, we need to change only once and don't have to grep in the code to find all the places to change.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -973,13 +978,16 @@ QByteArray Daemon::ApiImpl::SecretsRequestQueue::saltData() const
...
-        qCWarning(lcSailfishSecretsDaemon) << "Can't read salt data, assuming it's corrupted. DataProtector returned:" << s;
-        qCWarning(lcSailfishSecretsDaemon) << "NOTE: If you are running a pre-release sailfish-secrets version, you need to delete all old data before proceeding";
 
-        // TODO: find an appropriate solution for dealing with data corruption.
+    if (s == DataProtector::Irretrievable) {
+        qCWarning(lcSailfishSecretsDaemon) << "saltData: salt data is irretrieveably corrupted.";
+        dealWithDataCorruption();
+    } else if (s != DataProtector::Success) {
+        qCWarning(lcSailfishSecretsDaemon) << "saltData: can't read salt data. DataProtector returned:" << s;
+        // TODO: what should we do here? The data is not corrupted, but we can't access it.
```
dcaliste: Maybe we can notify the user here also, like for the data corruption case. In that case, the dealWithDataSomething() can be generalized to notify always and remove the directories only when data is corrupted.

daemon/SecretsImpl/secrets.cpp
Venemo: Just a typo, will fix.

daemon/SecretsImpl/secrets.cpp
Venemo: Maybe... I'll think about this some more.

chriadam:
First step is I think to ask Martin what the appropriate way to handle these cases are, from the user's POV.  E.g. how we should notify them, what sort of UI flow we provide to allow them to attempt to rectify the issue (e.g. re-enter master lock code to regenerate the test cipher data if that one got corrupted) etc.

For the salt data case, I agree that a corruption there is non-recoverable, so deleting the directory is probably the best option.  But I wonder if we should "wait" until the user has been informed and "confirms" the deletion action, instead of doing it immediately.  E.g. maybe in the future we will allow the salt data to be backed up in device backups, so they could then recover the salt data from such a backup, in order to continue using their existing secrets data.

Venemo:
@chriadam After talking with @rainemak about this, we came to the conclusion that this is a very very very rare occourence. So (for now at least) my suggestion is this:

* If we are afraid that the data corruption can actually happen, I suggest increasing the number of backup files from 3 to 5.
* Otherwise this PR is just about making sure that the user can continue to use their device without manually needing to drop into the terminal and deleting the corrupted secrets data.
* Since (ideally) this is something end user should never see, we shouldn't spend too much time on polishing it.

All that being said, I don't disagree with polishing this even further in the future, but considering the rarity of the addressed situation my humble opinion is that our time and resources could be better spent elsewhere.

chriadam:
Why is a separate request required?  Upon startup, the daemon will know whether the data is corrupted or not, and can store that internally.  Why not simply add an error code which uniquely identifies the issue, and return that for all requests?

daemon/SecretsImpl/secrets.cpp
```diff
@@ -2586,3 +2682,54 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
     }
 }
 
+void Daemon::ApiImpl::SecretsRequestQueue::dealWithDataCorruption() const
```
chriadam: Perhaps not necessary in the first implementation, but in the future it would be good if we dealt with salt data corruption differently to corruption of the test cipher data.  (In the latter case, we can recover by asking the user to re-enter the master lock code and regenerating the test cipher data.)  Perhaps add a parameter which describes the type of corruption (but is currently unused) would be useful, or add a comment regarding the possibility.

lib/Secrets/healthcheckrequest.cpp
```diff
@@ -0,0 +1,176 @@
...
+HealthCheckRequestPrivate::HealthCheckRequestPrivate()
+    : m_status(Request::Inactive)
+    , m_saltDataHealth(HealthCheckRequest::HealthUnknown)
+    , m_masterlockHealth(HealthCheckRequest::HealthUnknown)
+{
+}
+
+/*!
+ * \class HealthCheckRequest
+ * \brief Allows a client request information about the well being of secrets data.
```
chriadam: well-being should be hyphenated

lib/Secrets/healthcheckrequest.cpp
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * An example of retrieving information about available plugins follows:
```
chriadam: This line should be "... about the health of the secrets service data follows:"

lib/Secrets/healthcheckrequest.cpp
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * \code
+ * Sailfish::Secrets::SecretManager man;
+ * Sailfish::Secrets::HealthCheckRequest req;
+ * req.setManager(&man);
+ * req.startRequest();
+ * // status() will change to Finished when complete
+ * // real clients should not use waitForFinished().
+ * req.waitForFinished();
+ * qDebug() << "salt data health:" << req.
```
chriadam: unfinished

Venemo:
> Why is a separate request required?

You are not wrong, I could have chosen another approach. However when I started working on it the current approach seemed simpler and cleaner, that is why I went with it.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -2114,6 +2185,31 @@ void Daemon::ApiImpl::SecretsRequestQueue::handleFinishedRequest(
             }
             break;
         }
+        case GetHealthInfoRequest: {
```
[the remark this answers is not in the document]
Venemo: Yes, you are right. I wasn't 100% sure of this but was wondering myself if this is necessary.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -1223,6 +1261,39 @@ void Daemon::ApiImpl::SecretsRequestQueue::handlePendingRequest(
...
```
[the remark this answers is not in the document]
Venemo: The reason why it's here is because the static variables it uses are in this file. But if you are OK with duplicating those, then I can move it there.

lib/Secrets/healthcheckrequest.cpp
Venemo: Will fix!

lib/Secrets/healthcheckrequest.cpp
Venemo: This is just bad copy-paste, thanks for spotting it.

lib/Secrets/healthcheckrequest.cpp
Venemo: Will fix!

chriadam:
I'm not sure that it is simpler to increase the complexity of the API by adding an entire new request type, compared to adding two error codes, but perhaps I'm wrong.  How is the new request expected to be used - is every client expected to use that request prior to performing any other request, or is it only used if the daemon returns some specific error code to another type of request?

Are you able to prototype the "health error codes rather than health request" solution also for comparison?

chriadam:
After discussion on IRC I now agree with Venemo that adding a new request type is the correct solution, although for slightly different reasons.  While I think that it is simpler (from an API semantics perspective) to return the appropriate error code when required, instead of returning the more generic DatabaseLockedError and then requiring the client to perform the HealthCheckRequest to determine the precise issue, one very strong argument in favour of the HealthCheckRequest is that in the future we may need to extend the check to other things (checking individual plugins' metadata databases, for example) and thus the ability to extend the API and information reporting may be very useful.

daemon/SecretsImpl/secrets.cpp
Venemo: Honestly, I believe the occourence of this code ever running is so rare that I wouldn't bother with this. But added a comment now about the possibility of doing it in the future.

I would say if a client requests further work on it, let's do this. Otherwise let's not.

daemon/SecretsImpl/secrets.cpp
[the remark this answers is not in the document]
Venemo: Now removed the code from here, and just added a comment that explains why it isn't necessary there.

Venemo:
@chriadam I believe I've addressed most of your review findings now. Please go through it one more time and let me know if it's good now. Thank you :)

Venemo:
Added support for `HealthCheckRequest` to the tool. Now you can use:

```
secrets-tool --health-check
```

Using this I was able to test the feature and eliminate a couple of bugs in the code.

daemon/SecretsImpl/secrets.cpp
```diff
@@ -863,10 +897,12 @@ bool Daemon::ApiImpl::SecretsRequestQueue::determineTestCipherPlugin(
     DataProtector dataProtector(lockCodeCheckDirPath);
     DataProtector::Status s = dataProtector.getData(&previousData);
 
-    if (s != DataProtector::Success) {
-        qCWarning(lcSailfishSecretsDaemon) << "Can't read lock code data, assuming it's corrupted. DataProtector returned:" << s;
-        qCWarning(lcSailfishSecretsDaemon) << "NOTE: If you are running a pre-release sailfish-secrets version, you need to delete all old data before proceeding";
-        // TODO: find an appropriate solution for dealing with data corruption.
+    if (s == DataProtector::Irretrievable) {
+        qCWarning(lcSailfishSecretsDaemon) << "determineTestCipherPlugin: lock code data is irretrieveably corrupted.";
```
chriadam: typo, should be irretrievably

daemon/SecretsImpl/secrets.cpp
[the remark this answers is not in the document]
chriadam: Don't duplicate, just pass the secretsDirPath as a parameter, I think.

One other thing to check: which thread do we "normally" read the salt data etc within?  Is it always the main thread?  I assume so, but please double check this.

lib/Secrets/healthcheckrequest.cpp
```diff
@@ -0,0 +1,176 @@
...
+    , m_saltDataHealth(HealthCheckRequest::HealthUnknown)
+    , m_masterlockHealth(HealthCheckRequest::HealthUnknown)
+{
+}
+
+/*!
+ * \class HealthCheckRequest
+ * \brief Allows a client request information about the well-being of secrets data.
+ *
+ * An example of retrieving information about the secrets data health follows:
```
chriadam: Please add some more description to make it clear that this request is really intended for use by the system settings application only, and that most clients will never need to use this request.

lib/Secrets/secretmanager.cpp
```diff
@@ -110,6 +110,25 @@ SecretManagerPrivate::getPluginInfo()
...
 }
 
+
+QDBusPendingReply<Sailfish::Secrets::Result,
+                  HealthCheckRequest::Health,
+                  HealthCheckRequest::Health>
+SecretManagerPrivate::getHealthInfo()
+{
+    if (!m_interface) {
+        return QDBusPendingReply<Result>(
```
chriadam: this needs to return `QDBusPendingReply<Result, Health, Health>`.  I notice that the one above is also incorrect - can you please fix that one also?

daemon/SecretsImpl/secrets.cpp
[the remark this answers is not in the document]
Venemo: Done.

chriadam:
LGTM, thanks

PR #139, opened 2018-07-17
[sailfish-secrets] Add a helper constructIdenifier() function to be able to set it in DeleteStoredKeyRequest and StoredKeyRequest from QML.

dcaliste:
The Crypto::Key::Identifier is not exposed to QML, and even not a Q_GADGET. Following the wrapper for StoredKeyIdendifierRequest, I've added wrappers for DeleteStoredKey and StoredKey requests to be able to set the identifier.

### DIFF
--- a/lib/Crypto/deletestoredkeyrequest.h
+++ b/lib/Crypto/deletestoredkeyrequest.h
@@ -25,7 +25,7 @@ class DeleteStoredKeyRequestPrivate;
 class SAILFISH_CRYPTO_API DeleteStoredKeyRequest : public Sailfish::Crypto::Request
 {
     Q_OBJECT
-    Q_PROPERTY(Sailfish::Crypto::Key::Identifier identifier READ identifier NOTIFY identifierChanged)
+    Q_PROPERTY(Sailfish::Crypto::Key::Identifier identifier READ identifier WRITE setIdentifier NOTIFY identifierChanged)
 
 public:
     DeleteStoredKeyRequest(QObject *parent = Q_NULLPTR);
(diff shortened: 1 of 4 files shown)


dcaliste:
@chriadam or @Venemo I don't know if it's the better way to do it, but it works, I can call DeleteStoredKeyRequest from QML now.

dcaliste:
As an after thought, it may work with less code with something like that:
<pre>
Crypto::Key::Identifier CryptoManager::constructIdentifier(const QString &name, const QString &collectionName, const QString &storagePluginName)
{
  return Crypto::Key::Identifier(name, collectionName, storagePluginName);
}
</pre>
and from QML:
<pre>
CryptoManager {
  id: boss
}
DeleteStoredKeyRequest {
  id: eraser
  cryptoManager: boss
}
Button {
  onClicked: {
    eraser.identifier = boss.constructIdentifier("weak key", "example", "GnuPG")
    eraser.startRequest()
  }
}
</pre>
I'm doing something like that with FilterData in FindSecretsRequest from #140.

Venemo:
@dcaliste Feel free to move the code around. Note that not all non-QGADGETS are inaccessible from QML, so before giving my blessing on this I would like you to confirm that what you wrap is indeed otherwise inaccessible from QML. (Ie. doesn't it help to register the metatype or something?)

dcaliste:
@venemo, as commented before, I've tested the solution to make the identifier writable and add a constructIdentifier() method in the QML manager for the two delete and stored methods. It is working well and have minimal modifications.

Feel free to review the changes when you have time.

Venemo:
Looks good, thank you!

PR #140, opened 2018-07-25
Adjust various things to be able to use FindSecretsRequest from QML

dcaliste:
Adapt the code from StoredKeyIdentifiersRequest from Crypto to be able to list secret identifiers with FindSecretsRequest.

### DIFF
--- a/lib/Secrets/request.h
+++ b/lib/Secrets/request.h
@@ -22,7 +22,7 @@ class SAILFISH_SECRETS_API Request : public QObject
 {
     Q_OBJECT
     Q_PROPERTY(Sailfish::Secrets::SecretManager* manager READ manager WRITE setManager NOTIFY managerChanged)
-    Q_PROPERTY(Request::Status status READ status NOTIFY statusChanged)
+    Q_PROPERTY(Sailfish::Secrets::Request::Status status READ status NOTIFY statusChanged)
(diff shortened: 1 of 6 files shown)


dcaliste:
@chriadam or @Venemo, any remarks ?

Venemo:
@dcaliste I've looked through your changes. Looks good to me, but I don't see why exactly you needed to wrap the FindSecretsRequest, ie. what was wrong with it?

dcaliste:
Thanks for looking into this. It's exactly like Crypto/StoredKeyIdentifiersRequest. The identifiers output property is a list of Secrets::Secret::Identifier (_resp._ Crypto::Key::Identifier). But the Identifier class is not Q_GADGET and not exposable to QML (as far as I've understood it). Without the PR, a successfull request returning a list of two secrets for instance has a length shown as ```undefined``` from QML.

Maybe there is a better way to expose the identifiers property to QML, but I'm not an expert and just sheepishly followed what has been done for the crypto equivalent.

dcaliste:
It's fine with me. It's quite clear and it works transparently from QML.

dcaliste:
@venemo I've #139 that looks a bit the same, handling identifiers as input for some crypto requests, if you have time to look at ;)

Venemo:
@dcaliste Sure, will look at that too! :)

PR #142, opened 2018-08-31
Add more matrix tests

juniskan:


tests/scripts/matrix/run-matrix-tests.sh
```diff
@@ -9,23 +9,41 @@ fi
...
+           'n /home/nemo/manualtests/matrix/026.a.sh;n /home/nemo/manualtests/matrix/026.b.sh;n /home/nemo/manualtests/matrix/026.c.sh'
+           'n /home/nemo/manualtests/matrix/027.a.sh;n /home/nemo/manualtests/matrix/027.b.sh;n /home/nemo/manualtests/matrix/027.c.sh'
+           'n /home/nemo/manualtests/matrix/028.a.sh;n /home/nemo/manualtests/matrix/028.b.sh;n /home/nemo/manualtests/matrix/028.c.sh'
+           'p /home/nemo/manualtests/matrix/029.a.sh;p /home/nemo/manualtests/matrix/029.b.sh;p /home/nemo/manualtests/matrix/029.c.sh'
+           'p /home/nemo/manualtests/matrix/030.a.sh;p /home/nemo/manualtests/matrix/030.b.sh;p /home/nemo/manualtests/matrix/030.c.sh'
+           'p /home/nemo/manualtests/matrix/031.a.sh;p /home/nemo/manualtests/matrix/031.b.sh;p /home/nemo/manualtests/matrix/031.c.sh'
+           'p /home/nemo/manualtests/matrix/032.a.sh;p /home/nemo/manualtests/matrix/032.b.sh;p /home/nemo/manualtests/matrix/032.c.sh'
+	   'p /home/nemo/manualtests/matrix/033.a.sh;p /home/nemo/manualtests/matrix/033.b.sh;p /home/nemo/manualtests/matrix/033.c.sh'
```
chriadam: minor: whitespace here is tabspace, replace with spaces

chriadam:
Looking great - thanks very much for expanding these.  We should also add some rows to test removing the individual secret (or key) as well as removing the entire collection, as these are different codepaths which need testing.

Some of the crypto rows are still TODO, of course.  We should add some tests for crypto operations other than Sign and Verify, e.g. add GEDR (generate+encrypt+decrypt+remove) tests.

Also, we should add an --owner-only flag to the secrets-tool when creating a collection, to support testing OwnerOnly as well as NoAccessControl semantics via these tests (since currently secrets-tool --create-collection does setAccessControlMode(Sailfish::Secrets::SecretManager::NoAccessControlMode) always).

chriadam:
Thanks, I've merged into manualtests, and will merge from there into master shortly.  Nice work.

PR #143, opened 2018-09-04
[sailfish-secrets] Export lists to QML from PluginInfoRequest.

dcaliste:
As for #140, add wrappers for PluginInfoRequest so that the list attributes are available from QML.

@chriadam and @venemo, any remarks?


chriadam:
LGTM

dcaliste:
@chriadam do you think it can go in, or do you have any refrains on using it? Or it just went out of scope ;)

dcaliste:
@Venemo there is this one also, if you agree.

Venemo:
@dcaliste Merged. Have a nice day!

dcaliste:
@Venemo great, thank you.

PR #145, opened 2018-11-07
[sailfish-secrets] Call setFuture() in daemon after signal connection. Contributes to JB#36797

dcaliste:
There is a risk of race condition when connecting the `finished` signal after launching the thread with `setFuture()`. This is documented in the [Qt doc](http://doc.qt.io/qt-5/qfuturewatcher.html#setFuture).

I agree that the risk is quite neglectible here since the thread may have some work to do and may not return instantaneously, but it's safer by calling the setFuture() after the connection.


dcaliste:
Ok, corrected.

Issue #144, opened 2018-10-03
Add job oriented constructors or setters to PromptText


dcaliste:
Currently, to make the system dialog switches to CreatePassword state, it requires promptText to contain RepeatInstruction, see line passwordagentplugin.cpp#653. But IMHO, it's not possible to set RepeatInstruction with an empty string because of line 208 in interactionparameters.h. So to be in the CreateWindow state, one needs to give a string to the repeat instruction, which disable the possibility to use the default one.

We'd need some placeholder like "default" or something then if prompt.value(RepeatInstruction) == "default" then use the actual default one (or set it to empty etc to cause it to be the default one). it may be even more intuitive from API to have some prompt constructor doing this, like PromptText::PasswordCreationPrompt() ; or some flags to the constructor to specify the behaviour.

dcaliste:
After more thinking about this, I propose not to add a magic string acting like the default or add particular constructors in PromptText, but to modify the password plugin to detect password creation on operation() instead of relying on heuristic on PromptText content, see #146.

PR #146, opened 2018-11-15
[sailfish-secrets] Switch to password creation from operation(), and not from prompt texts. Contributes to JB#36797

dcaliste:
As discussed in #144 it is not possible at the moment to request a password creation input window without overriding the default repeat message.

This PR is fixing this by defining a new operation in InteractionParameters and detecting the creation request from operation in password plugin instead of relying on the prompt texts.

IMHO, it's more semantic to switch action depending on operation than trying to infer it from context (here prompt texts). Prompt texts should be kept for what they are: overriding default text messages in the system window.

### DIFF
--- a/lib/Secrets/interactionparameters.h
+++ b/lib/Secrets/interactionparameters.h
@@ -94,6 +94,8 @@ public:
         StoreKey            = 1 << 25,
         ImportKey           = 1 << 26,
 
+        CreatePassword      = 1 << 27,
+
         // reserved
         LastOperation       = 1 << 30
     };
--- a/plugins/gnupgplugin/gpgmebase.cpp
+++ b/plugins/gnupgplugin/gpgmebase.cpp
@@ -268,10 +268,7 @@ Result Daemon::Plugins::GnuPGPlugin::generateKey(const Key &keyTemplate,
         if (empty == kpgParams.customParameters().constEnd()) {
             Sailfish::Secrets::SecretManager secretManager;
             Sailfish::Secrets::InteractionParameters uiParams;
-            Sailfish::Secrets::InteractionParameters::PromptText prompt;
-            //% "Repeat password"
-            prompt.setRepeatInstruction(qtTrId("gnupg_plugin-repeat_passphrase"));
-            uiParams.setPromptText(prompt);
+            uiParams.setOperation(Sailfish::Secrets::InteractionParameters::CreatePassword);
             uiParams.setInputType(Sailfish::Secrets::InteractionParameters::AlphaNumericInput);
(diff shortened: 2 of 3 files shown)


dcaliste:
@chriadam what do you think?

@jpetrell this PR should solve one of the issues you reported about a missing string translation when creating a new GPG key and the passphrase window appears. Instead of supplying a string (that I forgot to add to translations) from the PGP plugin, we rely on the default ones.

lib/Secrets/interactionparameters.h
```diff
@@ -94,6 +94,8 @@ class SAILFISH_SECRETS_API InteractionParameters {
         StoreKey            = 1 << 25,
         ImportKey           = 1 << 26,
 
+        CreatePassword      = 1 << 27,
```
chriadam: Should this be "RequestPassword" rather than "CreatePassword"?

chriadam: Ah, no I understand now: this forces the repeat-prompt.  Ignore my comment above.

chriadam:
LGTM.  Can you please add "Contributes to JB#36797" to the commit subject?

chriadam:
Just double checking: all unit tests still pass?  Wondering if we relied on this prompt text for something in any test - I assume not.

chriadam:
I ran the tests on my device, and everything still passes.

Regarding Joona's earlier comment, this Operation value is interesting in that (as far as I can tell) it's mostly specified by the Daemon when talking to the Plugin, as opposed to something which is specified directly by the client.  Is that correct, or am I misunderstanding?

If that is the case, what happens in the case where the user specifies that operation directly and performs an interaction request?

dcaliste:
> this Operation value is interesting in that (as far as I can tell) it's mostly specified by the Daemon when talking to the Plugin, as opposed to something which is specified directly by the client. Is that correct, or am I misunderstanding?

I agree, but there is no specific request to create a password (in which case the daemon would properly place the operation flag by itself)). And I don't think it's necessary to add one. As far as I understand, this Operation enum is a _hint_ for the client of the interaction request. In my opinion, there should be a hint to say "I want an input from the user and will use the result as a new password", as we currently have for instance "i need input from the user and will use it to sign a document".

I think that up to now all operations were set from the daemon because we have a request for every operation _and_ the daemon is responsible to do an interaction request based on the choice in the request to be system only… But since we are exposing the InteractionRequest as a public API, we have no choice but to allow the user to set the operation hint. But, I agree that it's difficult to see it can have unexpected consequences.

chriadam:
LGTM.  We can address the latter concern through documentation in the future (e.g. document that that enum value has no effect if set by client code).  Thanks!

PR #147, opened 2018-11-22
[WiP] GObject API implementation for preliminary review

inzanity:


lib/SecretsCrypto/sf-secrets-collection.h
```diff
@@ -0,0 +1,68 @@
...
+	GObjectClass parent_class;
+};
+
+GType sf_secrets_collection_get_type(void);
+
+const gchar *sf_secrets_collection_get_plugin_name(SfSecretsCollection *collection);
+const gchar *sf_secrets_collection_get_encryption_plugin_name(SfSecretsCollection *collection);
+const gchar *sf_secrets_collection_get_name(SfSecretsCollection *collection);
+
+void sf_secrets_collection_get_secret(SfSecretsCollection *collection,
```
chriadam: In the C++ API, collections are exposed merely as data, while operations are performed via the Request APIs.  In this API, it seems like the collection is more "alive" (in that each instance somehow has a connection to the manager, and we have methods like these which get/set/delete secrets from a collection, passing in the collection itself rather than just its identifier).

What are the advantages of doing it this way, as opposed to keeping the API more symmetric / consistent with the C++ API?

lib/SecretsCrypto/sf-secrets-manager.c
```diff
@@ -0,0 +1,830 @@
...
+
+	(void)spec;
+
+	switch (property_id) {
+		case PROP_APPLICATION_ID:
+			if (priv->application_id)
+				g_free(priv->application_id);
+			priv->application_id = g_value_dup_string(value);
+			break;
+		case PROP_USER_INTERACTION_MODE:
```
chriadam: I wonder how many C applications will need to support in-app interaction mode?  I mean, if they provide a UI to allow the user to perform password entry or so on, I would assume that the UI would be a Sailfish.Silica / QML-based UI; and if that's the case, they should probably just use the C++ Secrets API.

I'm sure there are specific use cases for this (e.g. minimal UI case, some daemon which just talks to USB-key plugin), so it's probably needed eventually, but IMO getting everything else working first is probably higher priority.

lib/SecretsCrypto/sf-secrets-collection.h
inzanity: My though process was along the lines that a user of the API would be more likely to put several secrets in a single collection, it'd be a logical unit to have, and a logical split between source files too.

And also kind-of aiming for "if the collection object exists, the collection exists", although that cannot be guaranteed with the D-Bus API.

lib/SecretsCrypto/sf-secrets-manager.c
inzanity: I agree, this should've waited, I just wasn't familiar enough with the API yet when starting the implementation...

chriadam: Understandable, it's quite a complex domain.  If you ever have any questions about anything, just yell :-)

lib/SecretsCrypto/sf-secrets-collection.h
chriadam: Certainly I agree with the source code being split into different files.  I do understand the argument for the collection being the interaction endpoint (e.g. collection.save(secret), etc) from a "notional" standpoint, however there are (IMO) significant cognitive costs associated with that decision from an API perspective.  You noted one example of that above.

In general, my preference would be to mirror the C++ API as much as possible, but I am of course open to changing my mind if there are strong reasons to do otherwise in the C API context.

chriadam:
Purely cosmetic, but I don't know if we use short-form naming (like sf) anywhere, possibly "sailfish" could be more consistent even though more verbose.

inzanity:
The problem with long prefixes in C is that they are repeated. Everywhere. I'm not against s///g'in it entirely, but to me it makes sense to keep it short. Granted the function names etc are already long, and the added length doesn't add much, but IMO things are more readable when the common prefix is shorter.

monich:
I would consider shortening `sf_secrets_` prefix to just `sfs_` :)

e.g.
```
sfs_collection_get_encryption_plugin_name
sf_secrets_collection_get_encryption_plugin_name
sailfish_secrets_collection_get_encryption_plugin_name
```

inzanity:
I was thinking that too, not just for shortness, but also because then the _TYPE_ in macro names would sit nicely in more logical position. :)

lib/SecretsCrypto/sf-secrets-collection.h
inzanity: My goal was to write an API that I would personally like to use, but don't have strong opinion one way or another.

Changing it should be relatively low effort, much of the code can be used almost as-is.

monich: If you write an API that you personally like, if guarantees that somebody is going to like it (i.e. at least yourself). But if you write an API that even you don't like, it doesn't guarantee anything and it may end up being an API that no one likes!

lib/SecretsCrypto/crypto.h
```diff
@@ -1 +1,565 @@
...
+ */
+
+#include <stddef.h>
+#include <time.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+enum Sailfish_Crypto_Algorithm {
```
Venemo: Is there a better aproach than just copy-pasting the enums here? To me this looks like code duplication, and makes this error-prone.

lib/SecretsCrypto/sf-secrets-collection.h
Venemo: +1 for consistency with the C++ API.

lib/SecretsCrypto/sf-secrets-manager.c
Venemo: In principle a C app could also do some sort of in-app authentication, maybe in CLI rather than GUI?

lib/SecretsCrypto/sf-secrets.h
```diff
@@ -31,6 +36,119 @@ typedef enum SfSecretsPluginState_ {
 	SF_SECRETS_PLUGIN_STATE_PLUGIN_SUPPORTS_SET_LOCK_CODE = 1 << 2,
 } SfSecretsPluginState;
 
+typedef enum SfSecretsError_ {
```
Venemo: Could we somehow avoid duplicating the values like this?

lib/SecretsCrypto/crypto.h
inzanity: This file was not really supposed to be here anymore...

lib/SecretsCrypto/sf-secrets.h
inzanity: Techincally we could share the enums with C++ with a little moving around, but as C does not have namespaces, it becomes problematic. And also quite commonly enum values in C are written in UPPER_WITH_UNDERSCORES (in fact a requirement if GObject introspection is wanted).

What could be done is some preprocessing script that mangles the names from the C++ headers.

monich: IMO it's not worth the trouble - since these values are part of the D-Bus interface, they are not going to change in the future. Once synchronized, these values will never go out of sync.

chriadam: The DBus interface can change freely, no stability guarantees are provided for it.  But the values defined in the C++ header will not, as I intend to provide BC/SC guarantees for the C++ client library.  In any case, that doesn't change the fact that you're right: they're not going to change.

lib/SecretsCrypto/sf-secrets-manager.c
chriadam: Very true, and it does need to be done eventually for completeness, but I just think it's lower priority than e.g. getting the crypto API finished.

lib/SecretsCrypto/sf-crypto-key.c
```diff
@@ -0,0 +1,813 @@
...
+	return res;
+}
+
+GVariant *_sf_crypto_key_to_variant(SfCryptoKey *key)
+{
+	SfCryptoKeyPrivate *priv = sf_crypto_key_get_instance_private(key);
+
+	return g_variant_new("((sss)iiiiiayayaya(ay)a{sv})",
+			priv->name ?: "",
+			priv->collection_name ?: "",
```
chriadam: Seems unlikely, but if we switch to Clang then I'm not sure the elvis operator will continue to work, I think it might be gnu-extension

chriadam:
The Crypto API is looking good, really nice work.  Obviously a few more things to go (e.g. changing the secrets API a bit, supporting (symmetric) key derivation and (asymmetric) key generation parameters in the crypto API, and unit tests for everything, but definitely looking great!

rainemak:
What's the motivation behind variant to bool change in the dbus interface? Pondering who much we'll break with this?

rainemak:
That being said once commits are cleaned up we should get this integrated.

monich:
> Usually we use 4spaces per indentation level, now there are tabs. Could you replace those.

Let me leave my 2¢ here. I don't like tabs at all and generally try to avoid them but at the same time I don't feel good about forcing people to use spaces for indentation. As long as it's consistent within the component (i.e. tabs are not mixed with spaces) I would leave it the way it is in the form which is most comfortable to the person who is most likely going to maintain this code. Programming is a lot more fun if you like the code that you're writing which (hopefully) translates into quality. And the opposite doesn't translate into quality, that's for sure :)

chriadam:
What happened to cause all of the conflicts here?

rainemak:
@chriadam see https://github.com/sailfishos/sailfish-secrets/pull/151 . @inzanity prepared a cleanup commit set.

inzanity:
Obsoleted by #151

PR #149, opened 2018-12-11
[sailfish-secrets] Remove QtGui linkage. Contributes to JB#36025

chriadam:


jpetrell:
LGTM

dcaliste:
Fine, just a simple question: why is it required in qml/Secrets/Secrets.pro? I was thinking that Qt+=quick was enough for QQuickItem.

Venemo:
Instead of `QT -= gui` why not `QT = ...` ?

chriadam:
@dcaliste quick implies gui, so not strictly required, but also not harmful
@Venemo in case some included .pri conditionally adds support for something, better to use -= to avoid clobbering entire dependency list

adenexter:
> in case some included .pri conditionally adds support for something, better to use -= to avoid clobbering entire dependency list

That's why you put the QT = statement before any includes. If one of those .pri files adds gui you're still clobbering that but only one instance of two assuming it was added with += and not *=.

chriadam:
Fair point.  I've now rearranged those .pros so that the Qt modules are defined prior to any pri inclusion.

PR #150, opened 2018-12-12
[sailfish-secrets] Don't fetch cache passphrase on retry. Contributes to JB#43783

dcaliste:
Passphrases can be cached as Secrets, so they are not asked everytime. Before asking to enter a passphrase to unlock a GnuPG key, the pinentry is requesting access to a Secret collection to retrieve the passhrase if stored.

This commit disables this cache retrieval if the provided passphrase was erroneous (having an error generates a prompt instruction explaining about the error).

@chriadam may you review ?

### DIFF
--- a/plugins/gnupgplugin/pinentry/qassuanserver.cpp
+++ b/plugins/gnupgplugin/pinentry/qassuanserver.cpp
@@ -357,7 +357,7 @@ Sailfish::Secrets::Result QAssuanServer::requestPassphrase(QByteArray *passphras
     passphrase->clear();
 
     bool useCache = m_useCache->value(QVariant(true)).toBool();
-    if (useCache && cacheId.isValid()) {
+    if (useCache && cacheId.isValid() && prompt.instruction().isEmpty()) {
         Sailfish::Secrets::StoredSecretRequest request;
         qCDebug(lcSailfishPinentry) << "Starting cache request for" << cacheId.name();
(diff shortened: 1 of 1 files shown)


chriadam:
It seems somehow suboptimal to use the "prompt is empty" to ascertain that no error occurred during the passphrase step (as that seems like a "side effect").  Could the error value itself be passed somehow to that function?

If not, I am not against merging this as-is, but thought I'd raise the point above for discussion.

dcaliste:
GnuPG is passing information to the pinentry as strings, one by one. The C++ object QAssuanServer we have defined is storing these strings internally and when the ask_password command is received, it is creating the appropriated dialog with the stored strings.

Since the prompt object is itself storing strings, I decided to use it instead of a bunch of QStrings as private members to directly store the strings sent by GnuPG. There is (currently) no error related string in prompt, so I used instruction instead to store the error string (GnuPG is not using any instruction semantic as far as I know). I can document this better if you want.

Besides, if you look how Gnome or KDE are handling the GnuPG dialogs (type gpg -s foo to sign a file 'foo' to have a password dialog and enter a wrong passphrase), having actually an error in red colour or something could be quite nice to have: some possible later evolution in lipstick-security-ui ?

chriadam:
Ok.  Yes, there are a few improvements we could do for lipstick-security-ui, certainly.

PR #152, opened 2018-12-20
Add error handling tests

inzanity:
Add tests that cause errors to test error handling functionality.


chriadam:
LGTM, although I would prefer if Slava or someone else with more glib experience than I could review also.

lib/SecretsCrypto/sf-crypto.h
```diff
@@ -4,6 +4,9 @@
 #include <glib.h>
 #include <glib-object.h>
 
+#define SF_CRYPTO_ERROR (sf_crypto_error_quark())
+GQuark sf_crypto_error_quark(void);
```
chriadam: Is this a mapping between specific error enum values and some developer-readable string?

chriadam: It's important to note that in some cases, Crypto errors can include a reference to a Secrets error.  e.g. if a key comes from some storage plugin, and the key cannot be retrieved from the storage plugin for some reason, the crypto API will return an error which contains the crypto error code (e.g. unable to retrieve key) and the secrets error code (e.g. the caller is not allowed to access that collection).

inzanity: Changed to use the actual (more descriptive) storage error code in the SF_SECRETS_ERROR domain. Adding a mental note to document that errors from both domains are to expected when using key references.

PR #153, opened 2019-01-04
Add support for cipher sessions in the GObject API

inzanity:
This PR has two versions for update() and finish(), one that always returns a GBytes instance on success (the first commit), and in the next commit returns a boolean and uses NULL instead of empty GBytes.

I tend to like that the latter is somewhat less cumbersome, but both have their drawbacks. Comments on this welcome.

### DIFF
--- a/lib/SecretsCrypto/sf-common.c
+++ b/lib/SecretsCrypto/sf-common.c
@@ -10,6 +10,17 @@ GVariant *_sf_variant_new_bytes_or_empty(GBytes *bytes)
             sizeof(guchar));
 }
 
+GBytes *_sf_bytes_new_from_variant(GVariant *variant)
+{
+    gconstpointer data;
+    gsize n_elements;
+
+    data = g_variant_get_fixed_array(variant, &n_elements, sizeof(guchar));
+    return g_bytes_new_with_free_func(
+            data, n_elements * sizeof(guchar),
+            (GDestroyNotify)g_variant_unref, g_variant_ref(variant));
+}
+
 GBytes *_sf_bytes_new_from_variant_or_null(GVariant *variant)
 {
     gconstpointer data;
@@ -18,7 +29,9 @@ GBytes *_sf_bytes_new_from_variant_or_null(GVariant *variant)
     data = g_variant_get_fixed_array(variant, &n_elements, sizeof(guchar));
     if (n_elements == 0)
         return NULL;
-    return g_bytes_new(data, n_elements * sizeof(guchar));
+    return g_bytes_new_with_free_func(
+            data, n_elements * sizeof(guchar),
+            (GDestroyNotify)g_variant_unref, g_variant_ref(variant));
 }
 
 GVariant *_sf_variant_new_variant_map_string_or_empty(GHashTable *hash_table)
@@ -60,4 +73,22 @@ GHashTable *_sf_hash_table_new_string_from_variant(GVariant *variant)
     } while (g_variant_iter_next(&iter, "{sv}", &key, &item));
 
     return res;
+}
+
+GVariant *_sf_variant_new_variant_map_or_empty(GHashTable *hash_table)
+{
+    GVariantDict var_dict;
+
+    g_variant_dict_init(&var_dict, NULL);
+
+    if (hash_table) {
+        GHashTableIter ht_iter;
+        gpointer key;
+        gpointer value;
+        g_hash_table_iter_init(&ht_iter, hash_table);
+        while (g_hash_table_iter_next(&ht_iter, &key, &value))
+            g_variant_dict_insert_value(&var_dict, key, value);
+    }
+
+    return g_variant_dict_end(&var_dict);
 }
--- a/lib/SecretsCrypto/sf-crypto-cipher-session.h
+++ b/lib/SecretsCrypto/sf-crypto-cipher-session.h
@@ -1,0 +1,71 @@
+#ifndef SF_CRYPTO_CIPHER_SESSION_H
+#define SF_CRYPTO_CIPHER_SESSION_H
+
+#include <glib-object.h>
+#include <gio/gio.h>
+#include "sf-crypto-manager.h"
+#include "sf-crypto.h"
+#include "sf-crypto-key.h"
+
+#define SF_TYPE_CRYPTO_CIPHER_SESSION (sf_crypto_cipher_session_get_type())
+#define SF_CRYPTO_CIPHER_SESSION(o) (G_TYPE_CHECK_INSTANCE_CAST((o), SF_TYPE_CRYPTO_CIPHER_SESSION, SfCryptoCipherSession))
+#define SF_IS_CRYPTO_CIPHER_SESSION(o) (G_TYPE_CHECK_INSTANCE_TYPE((o), SF_TYPE_CRYPTO_CIPHER_SESSION))
+
+typedef struct SfCryptoCipherSession_ SfCryptoCipherSession;
+typedef struct SfCryptoCipherSessionClass_ SfCryptoCipherSessionClass;
+
+struct SfCryptoCipherSession_ {
+    GObject parent;
+};
+
+struct SfCryptoCipherSessionClass_ {
+    GObjectClass parent_class;
+};
+
+GType sf_crypto_cipher_session_get_type(void);
+
+void sf_crypto_cipher_session_new(SfCryptoManager *manager,
+        GBytes *iv,
+        SfCryptoKey *key,
+        SfCryptoOperation operation,
+        SfCryptoBlockMode block_mode,
+        SfCryptoEncryptionPadding padding,
+        SfCryptoSignaturePadding signature_padding,
+        SfCryptoDigest digest,
+        GHashTable *custom_parameters,
+        const gchar *provider_name,
+        GCancellable *cancellable,
+        GAsyncReadyCallback callback,
+        gpointer user_data);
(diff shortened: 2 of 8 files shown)


lib/SecretsCrypto/sf-crypto-cipher-session.h
```diff
@@ -0,0 +1,71 @@
...
+        SfCryptoBlockMode block_mode,
+        SfCryptoEncryptionPadding padding,
+        SfCryptoSignaturePadding signature_padding,
+        SfCryptoDigest digest,
+        GHashTable *custom_parameters,
+        const gchar *provider_name,
+        GCancellable *cancellable,
+        GAsyncReadyCallback callback,
+        gpointer user_data);
+SfCryptoCipherSession *sf_crypto_cipher_session_new_finish(GAsyncResult *res, GError **error);
```
monich: Are there synchronous versions of those too? In some cases (e.g. a command line app) synchronous calls are significantly more convenient.

inzanity: No synchronous versions are provided, wrappers could be provided with reasonable effort (some similarily as in the tests)

monich:
FWIW I would prefer `GBytes*` as a return value from *_finish calls (where appropriate), NULL indicating a failure.

lib/SecretsCrypto/sf-crypto-cipher-session.h
```diff
@@ -0,0 +1,71 @@
...
+        GBytes *authentication_data,
+        GHashTable *custom_parameters,
+        GCancellable *cancellable,
+        GAsyncReadyCallback callback,
+        gpointer user_data);
+gboolean sf_crypto_cipher_session_update_authentication_finish(GAsyncResult *res, GError **error);
+
+void sf_crypto_cipher_session_update(SfCryptoCipherSession *session,
+        GBytes *data,
+        GHashTable *custom_parameters,
```
monich: Can it possibly make sense to provide separate custom parameters for each update?

monich: To block-cipher serous amounts of data, I would expect an API providing pipe/socket descriptors to send the input and pull the output. Wrapping each block into a D-Bus transaction seems insane to me. Or can data contain more than one block? Or does this API only apply to asymmetric ciphers? Or any kind of cipher? Symmetric and asymmetric cases are significantly different and might deserve different APIs.

monich: Meh, it's probably just a reflection of the underlying D-Bus API, forget it...

lib/SecretsCrypto/sf-common.c
```diff
@@ -10,6 +10,15 @@ GVariant *_sf_variant_new_bytes_or_empty(GBytes *bytes)
             sizeof(guchar));
 }
 
+GBytes *_sf_bytes_new_from_variant(GVariant *variant)
+{
+    gconstpointer data;
+    gsize n_elements;
+
+    data = g_variant_get_fixed_array(variant, &n_elements, sizeof(guchar));
+    return g_bytes_new(data, n_elements * sizeof(guchar));
```
monich: Something like this would avoid copying the data:
```c
return g_bytes_new_with_free_func(data, n_elements,
    (GDestroyNotify) g_variant_unref, g_variant_ref(variant));
```

tests/SecretsCrypto/tst_secretscrypto/tst_secretscrypto.c
```diff
@@ -1567,6 +1569,449 @@ static void tst_crypto_import_invalid(SfCryptoFixture *fixture, gconstpointer te
...
+        g_assert_no_error(fixture->error);
+        g_assert_nonnull(random_bytes);
+
+        if (g_test_failed())
+            return;
+
+        g_byte_array_append(decoded, g_bytes_get_data(random_bytes, NULL), g_bytes_get_size(random_bytes));
+        g_bytes_unref(random_bytes);
+
+        g_assert_cmpmem(encoded->data, MIN(encoded->len, decoded->len),
```
chriadam: I'm not sure I understand what this comparison is doing.  I had thought that encoded->data should point to encrypted data, and decoded->data should point to decrypted data (i.e. identical to the sum of the random_bytes chunks), but if that were the case it looks like this comparison would ensure that the encrypted data is equal to the decrypted data, so I must be confused?

chriadam: Oh, never mind.  encoded->data is the sum of random_bytes chunks, not the encrypted data.  decoded->data is indeed the decrypted data, and thus they should indeed be equal.

tests/SecretsCrypto/tst_secretscrypto/tst_secretscrypto.c
```diff
@@ -1567,6 +1569,449 @@ static void tst_crypto_import_invalid(SfCryptoFixture *fixture, gconstpointer te
...
+
+        g_byte_array_append(encoded, g_bytes_get_data(random_bytes, NULL), g_bytes_get_size(random_bytes));
+
+        sf_crypto_cipher_session_update(encoder,
+                random_bytes,
+                NULL,
+                NULL,
+                _tst_crypto_ref_res_and_quit,
+                fixture);
+        g_bytes_unref(random_bytes);
```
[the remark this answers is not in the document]
inzanity: The problem here is that, unlike some other GLib data types, GBytes does not have a "floating reference" concept, and it is usually not nice to steal the callers reference (unless explicitly stated in the function name)

tests/SecretsCrypto/tst_secretscrypto/tst_secretscrypto.c
inzanity: This is a pretty useless comparison, the full comparison is done at the end of the function. At least with this test setup, the "decoded" array lags behind by some (IIRC 22) bytes. This just confirms that all data received thus far is valid.

tests/SecretsCrypto/tst_secretscrypto/tst_secretscrypto.c
```diff
@@ -1567,6 +1569,449 @@ static void tst_crypto_import_invalid(SfCryptoFixture *fixture, gconstpointer te
...
+        sf_crypto_cipher_session_update(decoder,
+                encoded_bytes,
+                NULL,
+                NULL,
+                _tst_crypto_ref_res_and_quit,
+                fixture);
+        g_bytes_unref(encoded_bytes);
+        g_main_loop_run(fixture->loop);
+
+        sf_crypto_cipher_session_update_finish(fixture->test_res, &random_bytes, &fixture->error);
```
chriadam: Are you able to also add a unit test which does multiple updates one-after-another without an update_finish in between each, to ensure that also works?  (I assume that update_finish waits for the dbus response?)

see e.g. https://github.com/sailfishos/sailfish-secrets/blob/master/tests/Crypto/tst_cryptorequests/tst_cryptorequests.cpp#L2103 for example of "batch" encryption mode with cipher session using C++ API.

tests/SecretsCrypto/tst_secretscrypto/tst_secretscrypto.c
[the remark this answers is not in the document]
chriadam: Fair enough, fine as-is then :-)

tests/SecretsCrypto/tst_secretscrypto/tst_secretscrypto.c
```diff
@@ -1567,6 +1569,449 @@ static void tst_crypto_import_invalid(SfCryptoFixture *fixture, gconstpointer te
...
+            NULL,
+            &verify_status,
+            &fixture->error);
+
+    g_assert_no_error(fixture->error);
+
+    if (g_test_failed())
+        return;
+
+    g_assert_cmpint(verify_status, ==, SF_CRYPTO_VERIFICATION_STATUS_SUCCEEDED);
```
chriadam: It might also be worthwhile to ensure that if you deliberately mess with random_bytes just prior to feeding them into the verifier, that verification fails, just to ensure that the API isn't just returning VERIFICATION_STATUS_SUCCEEDED for every input, etc.

chriadam: i.e. in a separate / second "expect failure" test.

chriadam:
I'm willing to be convinced otherwise, but I think returning gboolean from the _finish() functions is better, for two reasons: 1) the cipher session functions have to accommodate a variety of different operations including e.g. verify which doesn't produce any data as an output but only a verification status.  2) it's more consistent (both within the C API, and with the C++ api) to use the same return type from each function, IMO.

chriadam:
Regarding the file-descriptor passing, I agree that should be added eventually to the C API but only once it has been added to the C++ API and support for that is included in the daemon etc, and I guess that work will be done only if customers request that feature.

chriadam:
Mostly looks great to me.  I would like to see an expected-failure test added for verify at least, and also I think we should test that batch-mode cipher session updates work as expected without waiting for dbus round-trip.  But even those shouldn't block this PR from being merged I guess.

rainemak:
@inzanity, as @chriadam stated in most recent comment. Further changes could be done on top. I'd like to see this merged if this good in general.

chriadam:
The batch and expect-fail tests were added - thanks for that.  LGTM.

inzanity:
Rebased and cleaned history.

PR #154, opened 2019-02-19
[sailfish-secrets] Ensure that authentication plugin are installed with GnuPG plugins.

dcaliste:
Currently, pulling in the GnuPG plugin on a bare system will install the daemon but the authentication plugin that is used by pinentry is not installed.

@chriadam and @pvuorela this may solve some of the dependency issues we have went installing only this package on a bare system.


chriadam:
LGTM

Issue #155, opened 2019-03-07
InteractionRequest returns success on timeout


dcaliste:
Let `secrets-tool --get-user-input` run by itself, it is returning an empty string after a while. If you run it again and press the cancel button, it returns an error, detailing that operation has been canceled.

In my opinion, the timeout should be treated like a cancellation, at least. For instance, if the input request is used to inquire a password, letting it time out will validate the empty string as password…

More over, the dialog window is still shown while the console has already returned the empty string. We can even type some input there that are never returned anywhere.

@chriadam, I think there are two issues here:
* one with the secret daemon that timeout and returns the empty string;
* one with lipstick-security-ui that is not notified about the daemon giving up and keeping the window opened.

dcaliste:
As explained in #156, this is due to the fact that the password dialog itself timeout after the request. In the current implementation, dialog times out within a hard coded 5 minutes, while the request itself will time out within 2 minutes.

dcaliste:
Second issue (lipstick-security-ui that is not notified about the daemon giving up) is corrected by #158 .

dcaliste:
First issue (secret daemon timeouting is returning the empty string) should be solved by #158

PR #156, opened 2019-03-08
[sailfish-secrets] Ensure user input dialog timeout before request. Contributes to JB#44199

dcaliste:
Introduce a timeout attribute to interaction parameters. Use this
timeout value in password plugin to ensure that dialog is dismissed
before the request itself timeout. This avoid the request to return
an empty user input with a successful result while the dialog itself
is still being displayed.

Solve issue #155 

@chriadam there is still a possible issue with inconsistent hard coded timeout values: in `secretdaemonconnection.cpp:createInterface()` (/resp./ in crypto) the timeout is fixed to 3 minutes, which is fine for the default timeout of 2 minutes, but can create issues in my opinion if user sets up a longer timeout. The problem is that this timeout is set for the manager itself, so for all requests. I don't think it can be a good idea to change the timeout after creation when we want to manage a user input with a longer timeout.
A possible solution would be to fix timeouts as a constant somewhere in secret API and use this constant in password plugin. What do you think ?

### DIFF
--- a/plugins/passwordagentauthplugin/passwordagentplugin.cpp
+++ b/plugins/passwordagentauthplugin/passwordagentplugin.cpp
@@ -654,12 +654,16 @@ Result PasswordAgentPlugin::beginUserInputInteraction(
                                  was introduced. */
                               || promptText.contains(InteractionParameters::RepeatInstruction));
 
+    // SecretManager connection timeout is fixed at 3 minutes.
+    // Keep the timeout of the password dialog below this
+    // threshold to avoid having the dialog stay on screen
+    // while the request has already timeouted.
     QDBusPendingCall call = agent->asyncCall(newPassword
                     ? QStringLiteral("CreatePassword")
                     : QStringLiteral("QueryPassword"), {
                 QVariant::fromValue(cookie),
                 QVariant::fromValue(promptText.message()),
-                QVariant::fromValue(properties) }, 5 * 60 * 1000);
+                QVariant::fromValue(properties) }, 2 * 60 * 1000 + 30 * 1000);
 
     PasswordResponse * const response = new PasswordResponse(
                 call,


inzanity:
Not commenting on whether this is a feasible change or not, but do note that added properties are not automatically passed over D-Bus, they need to be changed in serialization.cpp; and such changes should be reflected in D-Bus interface introspection data, and any changes there should also be adapted onto the C API side.

If the timeout should only be passed between daemon and plugins, it is not necessary to touch the D-Bus interfaces.

chriadam:
I think specifying the timeout which the manager uses as some MaxTimeout value and exposing that in the API, makes sense.  Then adding the timeout()/setTimeout() for the user makes sense, so long as we document that values over MaxTimeout will be ignored (e.g. have setTimeout(int seconds) { m_timeout = qMin(MaxTimeout, seconds); }

Also as inzanity mentioned, will need to update the C API etc.

dcaliste:
Thank you @inzanity for your reminder. I've updated the first commit accordingly. Tell me if that's enough, particularly from the C API. It seems to me that InteractionParameters class is not wraped directly, but is just present in the xml file describing the API for DBus. I've added a `i` there.

@chriadam I'm proposing a second commit on top of the initial one, modifying the timeout setting per request in the Secret manager. I'm not convinced that it's a good choice. For requests without user interaction, I've used the default DBus value, for request with explicit interaction parameters, I've used the timeout chosen by the user. For the other requests that may use a built-in interaction parameter (read not set up by the user, but hard coded in the daemon), I'm using the default timeout value.

This is not translated into Crypto part. Indeed, trying to transfer it, I encounter the fact that InteractionParameters is duplicated (I know why), and didn't want necessary to transpose all Secret changes without discussion. And particularly to the fact that, even without actual user interaction, action may take time, more than the DBus default of 25 seconds, e.g. generation of random data. And some actions that don't take InteractionParameters as arguments may indeed create one from implementation. I'm thinking to the sign operation. In normal operation, it should not require a user interaction to unlock the key, because this is done in the fetch of the key. But for GnuPG plugin, the fetch is not unlocking the key, so the sign operation will take input from user to unlock the key… So when reading the implementation in cryptomanager.cpp, I didn't see any way to know which operation will actually require to wait for user interaction. So letting timeout at a hardcoded value like defaultTimeout() may work…

…or not: this is not taking into account possible operation time. Worst case scenario:
* the user choose a timeout of 3min ;
* manager set the timeout to 3min and launch the DBus call ;
* the operation starts by generating random data and show the user interface at T0+2min ;
* the user is distracted and takes more than one minute to enter is passphrase.
The issue #155 is then happening. The request will timeout but the password dialog will stay there for an additional two minutes before timeouting itself…

I don't know if you're still following me, but I'm afraid, correcting #155 is more complicated than just synchronizing timeouts as done here. I currently don't see how the client can signal the daemon, that OK, I give up, trash the request and clean pending operations.

inzanity:
The xml change is correct, but you'll need to add similar change in the second occurrence too, and the embedded xml in daemon/SecretsImpl/secrets_p.h and lib/Secrets/interactionservice_p.h. Also the C implementation needs to be changed, adding the i to the signatures where appropriate, and a value for it. Personally I'd suggest adding a `sf_secrets_manager_set_user_interaction_timeout()` and use the value wherever needed.

dcaliste:
As discussed with @chriadam on IRC today, we've decided to postpone the addition of client-side timeout. @inzanity thank you very much for detailed explanations, I'll use it nonetheless on a later MR.

This MR is now, just a quick and dirty fix for #155 ensuring that the password dialog timeout is lower than the request timeout.

I'm investigating with @chriadam advices how the daemon can notify the password plugin that the client has disconnected already and withdraw the dialog accordingly. Some news to come later today or this week in a new MR.

Then, I can reintroduce this timeout attribute from client-side in a third MR.

chriadam:
LGTM.  Merging.

PR #158, opened 2019-04-30
[sailfish-secrets] Add a cancellation mechanism to authentication and user input requests.

### LINKED ISSUE #157: Authentication dialogs are not cancelled when pinentry is aborted
Run `touch foo && gpg2 -s foo`, then abort it via `^C`. The current authentication dialog, either the password one, or the authorisation are not cancelled. It is related to #155, but is also an issue directly in pinentry implementation, since all request are blocking the Assuan loop.

The authentication requests should be done asynchronous to allow to catch assuan cancellation event.

dcaliste:
Closes #157 

Add listener in daemon on client DBus disconnect signal. Then notify the plugins to abort user input or authentication. Add ad hoc interface to plugins to handle cancellation.

### DIFF
--- a/daemon/CryptoImpl/crypto_p.h
+++ b/daemon/CryptoImpl/crypto_p.h
@@ -45,8 +45,7 @@ namespace Daemon {
 
 namespace ApiImpl {
 
-class CryptoRequestQueue;
-class CryptoDBusObject : public QObject, protected QDBusContext
+class CryptoDBusObject : public Secrets::Daemon::ApiImpl::DBusObject
 {
     Q_OBJECT
     Q_CLASSINFO("D-Bus Interface", "org.sailfishos.crypto")
@@ -561,9 +560,6 @@ public Q_SLOTS:
             const Sailfish::Crypto::InteractionParameters &interactionParameters,
             const QDBusMessage &message,
             Sailfish::Crypto::Result &result);
-
-private:
-    Sailfish::Crypto::Daemon::ApiImpl::CryptoRequestQueue *m_requestQueue;
 };
 
 class RequestProcessor;
@@ -586,6 +582,7 @@ public:
     bool unlockPlugin(const QString &pluginName, const QByteArray &lockCode);
     bool setLockCodePlugin(const QString &pluginName, const QByteArray &oldCode, const QByteArray &newCode);
 
+    void handleCancelation(Sailfish::Secrets::Daemon::ApiImpl::RequestQueue::RequestData *request) Q_DECL_OVERRIDE;
     void handlePendingRequest(Sailfish::Secrets::Daemon::ApiImpl::RequestQueue::RequestData *request, bool *completed) Q_DECL_OVERRIDE;
     void handleFinishedRequest(Sailfish::Secrets::Daemon::ApiImpl::RequestQueue::RequestData *request, bool *completed) Q_DECL_OVERRIDE;
     QString requestTypeToString(int type) const Q_DECL_OVERRIDE;
--- a/daemon/SecretsImpl/secretsrequestprocessor.cpp
+++ b/daemon/SecretsImpl/secretsrequestprocessor.cpp
@@ -1367,7 +1367,7 @@ Daemon::ApiImpl::RequestProcessor::userInput(
                                  callerPid,
                                  requestId,
                                  Daemon::ApiImpl::UserInputRequest,
-                                 QVariantList() << QVariant::fromValue<InteractionParameters>(promptParams)));
+                                 QVariantList() << QVariant::fromValue<QString>(userInputPlugin)));
     return Result(Result::Pending);
 }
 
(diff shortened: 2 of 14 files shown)


dcaliste:
@chriadam I solved the discussion we had about moving or not the RequestData struct into QObject, by making a base class on Secret/CryptoDBusObject and using the DBus context of this class to get the connection that is being disconnected and compare it with the list of pending requests.

It is working quite nicely. I can abort any CLI action requiring user input with a ^C now.

In a later PR, I will add timeout parameter to Secret/CryptoManager now that the daemon is able to actually cancel the pending operation.

In a much later PR, it would be nice to be able to cancel current crypto operation, like key generation for instance. But that's another story…

daemon/CryptoImpl/crypto_p.h
```diff
@@ -586,6 +582,7 @@ class CryptoRequestQueue : public Sailfish::Secrets::Daemon::ApiImpl::RequestQue
     bool unlockPlugin(const QString &pluginName, const QByteArray &lockCode);
     bool setLockCodePlugin(const QString &pluginName, const QByteArray &oldCode, const QByteArray &newCode);
 
+    void handleCancelation(Sailfish::Secrets::Daemon::ApiImpl::RequestQueue::RequestData *request) Q_DECL_OVERRIDE;
```
chriadam: maybe handleCanceledRequest() for symmetry with handleFinishedRequest()?  or maybe not, if a canceled request then transitions to finished state.  hrm.

dcaliste: Indeed, cancelation is not a request. As implemented now, it's an external event that requires the request queue to update itself and possibly notify the plugin having current request pending to cancel them.

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -6394,3 +6394,46 @@ void Daemon::ApiImpl::RequestProcessor::authenticationCompleted(
...
+        case DeleteCollectionRequest:
+        case StoredKeyIdentifiersRequest:
+        case SetCollectionSecretRequest:
+        case GetCollectionSecretRequest:
+        case GetStandaloneSecretRequest:
+        case FindCollectionSecretsRequest:
+        case DeleteCollectionSecretRequest:
+        case UseCollectionKeyPreCheckRequest:
+        case SetCollectionKeyPreCheckRequest: {
+            // always use the system authentication plugin for device lock authentication requests.
```
chriadam: But if the collection is custom locked, they may not have used the system authentication plugin, unless I'm misunderstanding?

plugins/passwordagentauthplugin/passwordagentplugin.cpp
```diff
@@ -500,9 +526,12 @@ Result PasswordAgentPlugin::beginAuthentication(
     m_polkitResponses.insert(cookie, response);
 
     connect(response, &QDBusPendingCallWatcher::finished, this, [=](QDBusPendingCallWatcher *watcher) {
-        watcher->deleteLater();
-
-        m_polkitResponses.remove(cookie);
+        QHash<QString, PolkitResponse *>::Iterator it = m_polkitResponses.find(cookie);
+        if (it == m_polkitResponses.end()) {
```
chriadam: slight behaviour change here: no watcher->deleteLater() ?

plugins/passwordagentauthplugin/passwordagentplugin.cpp
```diff
@@ -586,6 +615,30 @@ void PasswordAgentPlugin::startDeviceLockAuthentication(
...
+}
+
+void PasswordAgentPlugin::cancelAuthentication(
+        uint callerPid,
+        qint64 requestId)
+{
+    QHash<QString, PolkitResponse *>::Iterator it = m_polkitResponses.begin();
+    while (it != m_polkitResponses.end()) {
+        if (it.value()->callerPid == callerPid
+            && it.value()->requestId == requestId) {
```
chriadam: style nitpick: double indent condition continuation lines, to visually separate from the entailed code block

chriadam:
Created JB#45807 to track this issue internally

chriadam:
Just double checking: manual test case for this one is: run "secret-tool --get-user-input" and Crtl-C it, check that the user input dialog disappears (due to the daemon noticing the client disconnection, and cancelling the user input request with the authentication plugin)?

dcaliste:
Yes, that's the procedure to check the effect of the PR.

daemon/SecretsImpl/secretsrequestprocessor.cpp
dcaliste: Well, grepping beginAuthentication calls in the daemon (which are the ones we want to cancel in this switch entries), return only calls from system authentication plug-in. See result of `grep -r beginAuthentication daemon/`. That's why, I copy-pasted the code from all these places to get the same plugin.

If in the future, we allow custom authentication, we will need to store the plugin name, as done for the user input request above.

plugins/passwordagentauthplugin/passwordagentplugin.cpp
dcaliste: Ok, done.

plugins/passwordagentauthplugin/passwordagentplugin.cpp
dcaliste: Well, in fact it has been displaced just after the find(). I need the find result to quickly return if the authentication has been cancelled already. Then, having the find result at hand, I found more logical to move the deleteLater after the test that the watcher object still exists in m_polkitResponses.

chriadam:
Thanks very much!

PR #159, opened 2019-04-30
[sailfish-secrets] Return an error to client on DBus connection error. Fix Github#155

### LINKED ISSUE #155: InteractionRequest returns success on timeout

dcaliste:
Close #155.

The case when the DBus reply is an error was not treated previously. So, if the daemon is too slow to answer (or die?) and the CryptoManager or the SecretManager timeout, the result was successful and the result was empty.

Hopefully, I've catched all requests ;)

@chriadam or @Venemo any remarks ?

### DIFF
--- a/lib/Crypto/cipherrequest.cpp
+++ b/lib/Crypto/cipherrequest.cpp
@@ -730,15 +730,20 @@ void CipherRequest::startRequest()
                             needsStEmit = true;
                             this->d_ptr->m_status = Request::Finished;
                         }
-                        this->d_ptr->m_result = reply.argumentAt<0>();
-                        if (this->d_ptr->m_result.code() == Result::Succeeded) {
-                            this->d_ptr->m_cipherSessionToken = 0;
-                        }
-                        this->d_ptr->m_generatedData = reply.argumentAt<1>();
                         bool needsVfEmit = false;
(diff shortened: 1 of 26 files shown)


chriadam:
Created JB#45807 to track this issue internally

chriadam:
Just double checking: manual test for this case is to run "secret-tool --get-user-input" and then manually `devel-su kill -9 PID_OF_SECRETSD`, and then check that the appropriate (dbus) error message is printed by the secret-tool (i.e. handling the dbus error reply)?

dcaliste:
Yes, that's it also. Another test case (more common, I would say), is the timeout case. Run the get input command and let the request connection with the daemon timeout. Without the patch, the return value is null string with success, while it should be an error result.

chriadam:
Thanks very much!

PR #162, opened 2019-06-13
Add simple example application

chriadam:


examples/apps/cryptoexample/README
```diff
@@ -0,0 +1,38 @@
...
+Some notes on the implementation:
+
+- By using OwnerOnlyMode access control mode, the secrets
+  daemon will not allow any other application to access
+  the collection of secrets.
+
+- By using DeviceLock protection, the secrets daemon doesn't
+  need to ask the user for the lock code to unlock the
+  collection when the application attempts to access it,
+  but will ask for permission if required (e.g. after reboot).
```
chriadam: This has the disadvantage that the user will be presented a system dialog to let the application access the secret, the next time the application is run after a reboot.  This behaviour (always require user to explicitly unlock collections) was requested by one of our partners, but it might be useful to add an "ApplicationData" mode which doesn't require any user input.

(This mode can currently be emulated by registering an application-specific InteractionView which always returns the same key, instead of actually prompting the user for input, however that is obviously a suboptimal workaround.)

rpm/sailfish-secrets.spec
```diff
@@ -400,6 +400,7 @@ ln -s ../sailfish-secretsd.service %{buildroot}/%{user_unitdir}/user-session.tar
 
 %files -n libsailfishcrypto-tests
 %defattr(-,root,root,-)
+%{_bindir}/sailfishcryptoexample
```
chriadam: it actually belongs in a separate examples package, but I'm not sure how much value there is in adding yet another package to this spec, for this example which is primarily for documentation purposes anyway...

chriadam:
Nobody had any concerns with this, so I will merge - it's useful for documentation purposes.

jpetrell:
Late LGTM :)

PR #164, opened 2019-07-25
[sailfish-secrets] Fix docs for C++ module indexes. Fixes JB#46721

martyone:


chriadam:
I think there should be a link to the sailfishcrypto-module.html page added to the sailfish-crypto-overview.html page (or is there one and I missed it?) and the same for the secrets one too.  Aside from that, LGTM!

martyone:
There isn't such a link yet, but as the overview pages already list the (main?) classes, and divides the listing by topics, I am not sure if it is really needed to link it there. For me it would be OK to leave it accessible just from the contents pane. I will merge this as is and you can add the links if you feel them really needed.

PR #166, opened 2020-01-16
[sailfish-secrets] Register types with appropriate meta type names. Contributes to JB#45215

chriadam:
This commit undoes some of the type registration naming done in
2d1f6f80 and further improves the namespacing and type registration
to ensure that clients can use the required types unambiguously.

This change is source incompatible for clients using the types from
QML, but sailfish-secrets is not allowed for harbour currently.

### DIFF
--- a/qml/Crypto/main.cpp
+++ b/qml/Crypto/main.cpp
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2018 Jolla Ltd.
+ * Copyright (C) 2018 - 2020 Jolla Ltd.
+ * Copyright (C) 2020 Open Mobile Platform LLC.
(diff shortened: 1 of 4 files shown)


dcaliste:
@chriadam thanks for clarifying these. It may break jolla-settings-accounts/extensions/ui/EmailCryptoKeySelection.qml where I use both Secret and Cryypto import without `as`-ing them, and I use also things like `SecretsRequest`.

I can test and propose a MR there to fix it.

chriadam:
@dcaliste thanks for pointing that one out!  created https://bitbucket.org/jolla/ui-jolla-settings-accounts/pull-requests/72/jolla-settings-accounts-adapt-to-updated/diff

blammit:
Tested with Settings > Keys UI changes, LGTM.

PR #167, opened 2020-01-31
[SecretsCrypto] Implement in-app authentication. Contributes to JB#44191

inzanity:
PR also includes changes in the test plugin and some portability fixes / not relying on undefined execution order in the daemon.


daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -6133,14 +6197,19 @@ Daemon::ApiImpl::RequestProcessor::userInputInteractionCompleted(
                         returnResult = Result(Result::UnknownError,
                                               QLatin1String("Internal error: incorrect parameter count!"));
                     } else {
+                        Secret::Identifier identifier = pr.parameters.takeFirst().value<Secret::Identifier>();
+                        Sailfish::Crypto::CryptoManager::Operation operation = pr.parameters.takeFirst().value<Sailfish::Crypto::CryptoManager::Operation>();
+                        QString cryptoPluginName = pr.parameters.takeFirst().value<QString>();
+                        CollectionMetadata collectionMetadata = pr.parameters.takeFirst().value<CollectionMetadata>();
+                        SecretManager::UserInteractionMode userInteractionMode = static_cast<SecretManager::UserInteractionMode>(pr.parameters.takeFirst().value<int>());
```
rainemak: Order of taking out parameters has changed. Intentional?

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -6150,12 +6219,15 @@ Daemon::ApiImpl::RequestProcessor::userInputInteractionCompleted(
                         returnResult = Result(Result::UnknownError,
                                               QLatin1String("Internal error: incorrect parameter count!"));
                     } else {
+                        Secret::Identifier identifier = pr.parameters.takeFirst().value<Secret::Identifier>();
+                        CollectionMetadata collectionMetadata = pr.parameters.takeFirst().value<CollectionMetadata>();
+                        SecretManager::UserInteractionMode userInteractionMode = static_cast<SecretManager::UserInteractionMode>(pr.parameters.takeFirst().value<int>());
```
rainemak: Same as above?

lib/SecretsCrypto/sf-secrets-interaction-request.c
```diff
@@ -77,29 +282,272 @@ static void _sf_secrets_interaction_request_get_property(GObject *object, guint
...
+    return priv->echo_mode;
+}
+
+static gboolean _sf_hash_table_equal(GHashTable *a, GHashTable *b, GEqualFunc value_equal)
+{
+    GHashTableIter i;
+    gpointer key;
+    gpointer value;
+
+    if (a == b)
```
rainemak: If both are null, then we're returning true which is kind of ok. Just wondering the order of branches and if it this should return false whenever either is null.

daemon/SecretsImpl/secretsrequestprocessor.cpp
inzanity: Good question, let's see...

inzanity: Yes, good catch. Still seems that there are places that push the same request with only 2 arguments, but not really relevant for this.

daemon/SecretsImpl/secretsrequestprocessor.cpp
inzanity: Yep.

lib/SecretsCrypto/sf-secrets-interaction-request.c
inzanity: I think this is quite reasonable as it is; an improvement could be to consider NULL and empty equal, but as code elsewhere tries to avoid creating empty HTs, should be good as-is.

rainemak:
LGTM

lib/SecretsCrypto/sf-secrets-manager.c
```diff
@@ -153,27 +182,66 @@ static gboolean _sf_secrets_manager_perform_interaction(SfSecretsManager *manage
...
+    gint32 operation;
+    const gchar *authentication_plugin_name;
+    GVariant *prompt_text;
+    GHashTable *pt;
+    gint32 input_type;
+    gint32 echo_mode;
+
+    g_variant_get(args,
+            "(&s&s&s&s(i)&s@a{is}(i)(i))",
+            &secret_name,
```
spiiroin: This ig_variant_get() s all new function to me... so, I'm wondering that apparently there is no return value and possible deconstruction hiccups can't be detected that way -> Is there any guarantee that all these uninitialized pointers actually get cleared / filled in before they are used later on?

lib/SecretsCrypto/sf-secrets-manager.c
```diff
@@ -153,27 +182,66 @@ static gboolean _sf_secrets_manager_perform_interaction(SfSecretsManager *manage
...
     g_object_add_weak_pointer(G_OBJECT(request), (gpointer *)&request);
     g_signal_emit(manager,
             _sf_secrets_manager_signals[SIGNAL_NEW_INTERACTION_REQUEST], 0,
             request,
             &result);
     g_object_unref(request);
 
-    if ((request = g_object_ref(request))) {
+    if (request) {
+        request = g_object_ref(request);
```
spiiroin: This explicitly let go of reference and then continue using it, even reacquiring a ref looks really odd (even if the weak ref stuff makes it work ok?). Why not just move the unref below if/else block and avoid murky waters?

lib/SecretsCrypto/sf-secrets-manager.c
```diff
@@ -204,17 +308,25 @@ static gboolean _sf_secrets_manager_cancel_interaction(SfSecretsManager *manager
...
 
     if (!request)
         return FALSE;
 
-    g_hash_table_remove(priv->requests, request_id);
-    g_object_weak_unref(G_OBJECT(request), (GWeakNotify)_sf_secrets_manager_remove_request, manager);
-    g_object_set(request,
-            "invocation", invocation,
-            NULL);
+    g_object_add_weak_pointer(G_OBJECT(request), (gpointer *)&request);
```
spiiroin: 'request' is in stack. Is it really ok to return and leave a weak ref to it behind?

lib/SecretsCrypto/sf-secrets-manager.c
```diff
@@ -274,8 +394,8 @@ static void _sf_secrets_manager_set_property(GObject *object, guint property_id,
                         g_direct_equal,
                         g_object_unref,
                         NULL);
-                priv->requests = g_hash_table_new(g_str_hash,
-                        g_str_equal);
+                priv->requests = g_hash_table_new_full(g_str_hash,
+                        g_str_equal, (GDestroyNotify)g_free, NULL);
```
spiiroin: g_free() should match GDestroyNotify -> unnecessary cast?

plugins/inappauthplugin/plugin.cpp
```diff
@@ -55,90 +55,91 @@ Daemon::Plugins::InAppPlugin::beginUserInputInteraction(
...
+                    QLatin1String("Enter the lock code for the unit test"))) {
+            QMetaObject::invokeMethod(this, "userInputInteractionCompleted", Qt::QueuedConnection,
+                                      Q_ARG(uint, callerPid),
+                                      Q_ARG(qint64, requestId),
+                                      Q_ARG(Sailfish::Secrets::InteractionParameters, interactionParameters),
+                                      Q_ARG(QString, interactionServiceAddress),
+                                      Q_ARG(Sailfish::Secrets::Result, Sailfish::Secrets::Result(Sailfish::Secrets::Result::Succeeded)),
+                                      Q_ARG(QByteArray, QByteArray("example lock code for the unit test")));
+            return Result(Result::Pending);
+        } else if (interactionParameters.promptText().instruction().endsWith(
```
spiiroin: qt style nit: Do not put 'else' after jump statements

tests/SecretsCrypto/tst_secretscrypto/tst_secretscrypto.c
```diff
@@ -725,6 +727,95 @@ static void tst_secret_create_existing_collection(SfSecretsFixture *fixture,
             SF_SECRETS_ERROR_INVALID_COLLECTION);
 }
 
+static void _tst_secret_interaction_request_continue(SfSecretsInteractionRequest *request,
+        SfSecretsFixture *fixture)
+{
+    GBytes *b;
+    (void)fixture;
+    (void)request;
```
spiiroin: 'request' is used below - or am I missing something?

lib/SecretsCrypto/sf-secrets-manager.c
inzanity: Good point, it is simply returns without doing anything if the string doesn't match the value. If guess some additional checks would be smort to avoid undefined things.

lib/SecretsCrypto/sf-secrets-manager.c
inzanity: The point here is to not run the block below, unless someone handling the "new interaction request" takes a ref on the object. It indeed looks kinda shady.

lib/SecretsCrypto/sf-secrets-manager.c
spiiroin: Well, initializing pointers to null would already make any (theoretical?) problems more likely to manifest themselves in deterministic manner / right context via immediate crash at null deref vs stack littering, or?

lib/SecretsCrypto/sf-secrets-manager.c
inzanity: Definitely not, this is a boog in the non-happy case.

spiiroin:
Missed that this was already merged. I guess that possible weak ref leakage is the only real issue I saw - if what I saw works the way I assume it works ;-)

lib/SecretsCrypto/sf-secrets-manager.c
inzanity: Well, yes, that would be one way of addressing it. (Obviously the wrong one <flamebait emoji>). I would prefer to add a g_variant_check_format_string call (maybe adding a helper that does _check and _get).

lib/SecretsCrypto/sf-secrets-manager.c
inzanity: True.

lib/SecretsCrypto/sf-secrets-manager.c
spiiroin: As far as I am concerned the only obviously wrong approach is to allow uninitialized pointers to be used as valid ones. The rest is subjective, depends on circumstances, and library of books have been written about the supposedly best ways to deal with it.

inzanity: Agreed.

daemon/SecretsImpl/secretsrequestprocessor.cpp
```diff
@@ -5890,17 +5890,25 @@ Daemon::ApiImpl::RequestProcessor::userInputInteractionCompleted(
                         returnResult = Result(Result::UnknownError,
                                               QLatin1String("Internal error: incorrect parameter count!"));
                     } else {
+                        QString collectionName = pr.parameters.takeFirst().value<QString>();
```
chriadam: these should be marked as const where possible

chriadam:
With current master (tagged version 0.2.17) I get a failure and a crash in the tst_secretsrequests unit test, whereas with previous tagged version (0.2.16) no such failure or crash occurs.

FAIL!  : tst_secretsrequests::lockCode() Compared values are not the same
   Actual   (lcr.result().code()): Failed
   Expected (Result::Succeeded)  : Succeeded
   Loc: [tst_secretsrequests.cpp(1638)]

inzanity:
Did you run a fresh test with 0.2.16, for me the `lockCode()` test fails with it too?

chriadam:
I thought I did, but might have messed up?  I am about to head home, but will check that tomorrow.  If it fails still with 0.2.16, I will fix it as part of JB#45000 assuming that 3.3.0 is still open for changes.

chriadam:
No, I just installed 0.2.16, rebooted, and retested again, and with 0.2.16 the tst_secretsrequests test has no failures.

```
Totals: 15 passed, 0 failed, 1 skipped, 0 blacklisted
********* Finished testing of tst_secretsrequests *********
```

inzanity:
Weird, I built from tag 0.2.16, installed and rebooted, and:
```
Totals: 13 passed, 2 failed, 1 skipped, 0 blacklisted
********* Finished testing of tst_secretsrequests *********
```

chriadam:
Hmm.  Did you delete the database each time?  Are you stopping the daemon and starting it in test mode as described in the README.md also?

inzanity:
Argh, I wasn't actually running the test in privileged mode, my bad.

inzanity:
Yes, this is my booboo in 85b3431e5a38015d812025eadac3d79614dbd0ed. The good news is that this only affects tests, the bad news is that I somehow missed this.

inzanity:
Hmm, the `SecretsCrypto` tests run just as well without that commit; I think it can be safely reverted.

PR #168, opened 2020-02-03
[SecretsCrypto] Plug weak reference leak. Contributes to JB#44191

inzanity:


spiiroin:
:+1: 

Side note: The issue is fixed, but I had to look twice due to slightly confusing surrounding code. First there is early return on null request. Then there is unconditional add weak pointer to requrest. Then the weak ref is removed on condition that request is non-null -> copy paste fluff from something that did more complex things???

inzanity:
The idea is that a well-behaving authentication provider would drop the reference to the `request` in the `cancel`-signal handler. This has certain drawbacks with things like threading, but seemed like a reasonable thing. The same principle can be extended further by adding a delay before sending the error reply, but as that can be done without breaking the API/ABI, this seemed like a safe and "simple" first implementation.

plugins/inappauthplugin/plugin.cpp
```diff
@@ -55,91 +55,90 @@ Daemon::Plugins::InAppPlugin::beginUserInputInteraction(
...
-            QMetaObject::invokeMethod(this, "userInputInteractionCompleted", Qt::QueuedConnection,
-                                      Q_ARG(uint, callerPid),
-                                      Q_ARG(qint64, requestId),
-                                      Q_ARG(Sailfish::Secrets::InteractionParameters, interactionParameters),
-                                      Q_ARG(QString, interactionServiceAddress),
-                                      Q_ARG(Sailfish::Secrets::Result, Sailfish::Secrets::Result(Sailfish::Secrets::Result::Succeeded)),
-                                      Q_ARG(QByteArray, QByteArray("sailfish"))); // passphrase for importKey unit tests
-            return Result(Result::Pending);
-        }
+    if (interactionServiceAddress.isEmpty()
```
chriadam: was this change intentional?  Without looking deeper, it looks like the original intention was that the test plugin behaviour would only occur if the interaction service address is empty.

chriadam: Oh, I see, this is a reversion of the breaking change; never mind.

chriadam:
LGTM, although I'm not familiar with the glib code

chriadam:
Is this one waiting for some more changes, or can it be merged?

inzanity:
I'd forgotten the whole thing, I think it can be merged since no objections have arisen.

PR #170, opened 2020-07-27
[SecretsCrypto] Update notification configuration. JB#50448

blammit:
- Set summary/body instead of previewSummary/previewBody.
- Add displayName to actions.


adenexter:
LGTM

PR #178, opened 2021-10-14
[sailfish-secrets] Put the P2P DBus socket in a sub-directory of /run. Contributes to JB#56300

dcaliste:
The DBus dialog is currently happening on `/run/user/100000/sailfishsecretsd-p2pSocket`. For instance when you need a confirmation to access the secret storage, or you require a password.

But for jailed application, /run is no readable. So like QMF, this MR is moving the socket to a subdirectory to allow to whitelist it.

@chriadam do you agree ?

### DIFF
--- a/daemon/controller.cpp
+++ b/daemon/controller.cpp
@@ -29,13 +29,13 @@ namespace {
             return QString();
         }
 
-        QDir dir(path);
-        if (!dir.mkpath(dir.absolutePath())) {
+        QDir dir(QDir(path).absoluteFilePath(QString::fromUtf8("sailfishsecretsd")));
+        if (!dir.mkpath(dir.path())) {
             qCWarning(lcSailfishSecretsDaemonDBus) << "Could not create socket file directory";
             return QString();
         }
 
-        const QString socketFile = QString::fromUtf8("%1/%2").arg(dir.absolutePath(), QLatin1String("sailfishsecretsd-p2pSocket"));
+        const QString socketFile = dir.absoluteFilePath(QLatin1String("p2pSocket"));
         const QString address = QString::fromUtf8("unix:path=%1").arg(socketFile);
 
         return address;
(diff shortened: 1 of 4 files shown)


Thaodan:
Technically user services should not really directly have services in /run but only in /run/user. What stops us from allowing writing to to this socket with a permissions?

dcaliste:
Sorry, it seems I'm still not clear.

When you're using gpgme library in your code, your code will dialog with the gpg-agent, using dedicated sockets allowed in GnuPG.permission file, as you mentioned.

Then, the gpg-agent may launch a pinentry program to get the passphrase from the user, for signature handling or decryption for instance.

Sailfish secret is providing a pinentry program (like Gnome is doing for instance) that is interfaced with Lipstick. To work, this pinentry program is discussing with the secret daemon over DBus to get a user input interaction. So the pinentry requires to access the DBus p2p socket established by the secret daemon.

At the moment, a gpg-agent is not permanently running in the user session. So one is launched on demand by a program using gpgme for instance. If the program using gpgme is running inside a jail, the gpg-agent will also run inside the jail. Thus it won't be able to see the secret daemon socket. Except if this socket is whitelisted. But as far as I now, we can only white list directories, and whitelisting `/run/user/100000` would be a breach in isolation.

The solution that was proposed in QMF for instance, was to move the socket from `/run/user/100000` to a dedicated subdirectroy, so this subdirectory only can be whitelisted for Email permission. I'm applying the same solution here for the secret daemon.

Actually the PR here has nothing to do with GnuPG. This PR is more about allowing permission for an application to talk to the secret daemon. At the moment, not any jailed application can (and particularly the gpg-agent from a jail), thus this PR.

Another completely independent solution to my problem of email signing not working would have been to run a gpg-agent continuously in the user session, or having it started by systemd, so it could always access the secret daemon socket.

chriadam:
Yes, it needs to be in its own directory, so that the sailjail whitelisting can work.  Thank you.  @Tomin1 can you check this one also, please?

Tomin1:
> Yes, it needs to be in its own directory, so that the sailjail whitelisting can work.

That's right.

> Thank you. @Tomin1 can you check this one also, please?

Looks fine to me too. :+1:

PR #180, opened 2021-11-22
[sailfish-secrets] Have license file as %license. JB#55991

pvuorela:
@Tomin1 @mlehtima


Tomin1 (approved):
This is fine, but I think `License` tag should be `BSD` instead of `BSD-3-Clause` like it is now.

mlehtima (approved):

pvuorela:
> This is fine, but I think License tag should be BSD instead of BSD-3-Clause like it is now.

Yes, true. I'll update as such and merge.

PR #182, opened 2021-12-02
Revert "[sailfish-secrets] Correct p2p socket name for the password a…

dcaliste:
…gent. Contributes to JB#56300"

This reverts commit 9e7cb2a22882f06df260adb90d8d651f340f5259.
The p2pSocket-agent is used by the password plugin to communicate
with lipstick. The p2pSocket is the daemon socket only.

@chriadam , I'm so sorry for such confusion. I don't know what went wrong when I tested the 4.4.0.1 version and get an error with the lipstick-secuirty-ui making the prompt window not appearing. I've recompiled the master branch reverting the erroneous commit I made and pushed to you. And it's now working properly with 4.4.0.12...


chriadam:
NP, LGTM.  I thought I tested the original version ;-)

PR #183, opened 2021-12-14
[sailfish-secrets] Create the socket directory in the plugin also. Contributes to JB#56300

dcaliste:
Plugin are initialised before the daemon is creating its own
socket and ensuring the exitence of the directory under
/run/user/xxx/sailfishsecretsd. So create also that directory
in the password agent plugin which is using this directory.

@chriadam : this is the true root of the problem, I guess. By moving the socket under a subdirectory, I created an issue with this plugin that was then unable to start its QDBusServer because the path was not existing yet (it's created later on by the daemon). It also explains why when testing it was always working : because then the directory was existing and everything was working after a daemon restart. While it was not working directly after a boot of the device.

### DIFF
--- a/plugins/passwordagentauthplugin/passwordagentplugin.cpp
+++ b/plugins/passwordagentauthplugin/passwordagentplugin.cpp
@@ -18,6 +18,7 @@
 #include <QtDBus/QDBusReply>
 
 #include <QtCore/QFile>
+#include <QtCore/QDir>
 #include <QtCore/QLoggingCategory>
 #include <QtCore/QTimer>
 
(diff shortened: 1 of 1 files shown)


chriadam:
Ah, interesting... is it not possible to force the daemon to create the appropriate sockets and socket directory, before initialising the plugins?

chriadam:
But I guess this is correct too, since this plugin is using that directory.  LGTM.

chriadam:
I will ask Jorma if this can sneak into the 4.4.0 release...

dcaliste:
> But I guess this is correct too, since this plugin is using that directory.

Yes, I prefer to create the directory where it is used, than assumed that it will be created at the right time by the daemon.

chriadam:
LGTM, tyvm.

dcaliste:
Great, thank you and merry Christmas.

PR #189, opened 2023-08-31
[sailfish-secrets] Properly detect the key ID in case of a subkey.

dcaliste:
Also don't bother the user with a second demand to access the cache passphrase if the first attempt was dismissed.

@pvuorela , if you have time to give a look. Thank you.

### DIFF
--- a/plugins/gnupgplugin/pinentry/qassuanserver.cpp
+++ b/plugins/gnupgplugin/pinentry/qassuanserver.cpp
@@ -77,15 +77,17 @@ gpg_error_t _assuan_cmd_setdesc(assuan_context_t ctx, char *line)
 {
     qCDebug(lcSailfishPinentry) << __func__ << line;
     QAssuanServer *self = static_cast<QAssuanServer*>(assuan_get_pointer(ctx));
-    self->prompt.setMessage(QByteArray::fromPercentEncoding(line));
+    self->prompt.setMessage(QByteArray::fromPercentEncoding(line).simplified());
 
     // Ugly hack here due to GnuPG being stuck in 2.0.4
-    // and not having SETKEYINFO command yet. Description may contain "ID xxxxx".
+    // and not having SETKEYINFO command yet. Description looks like:
+    // "user: "John Dow <john.dow@example.org>""
+    // "2048-bit RSA key, ID A099539B, created 2018-04-13 (main key ID 41123AA4)"
     {
         const QString &str(self->prompt.message());
-        int id = str.lastIndexOf("ID ");
+        int id = str.lastIndexOf(", ID ");
         if (id > 0) {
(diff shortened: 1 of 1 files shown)


plugins/gnupgplugin/pinentry/qassuanserver.cpp
```diff
@@ -77,15 +77,15 @@ gpg_error_t _assuan_cmd_setdesc(assuan_context_t ctx, char *line)
...
     QAssuanServer *self = static_cast<QAssuanServer*>(assuan_get_pointer(ctx));
-    self->prompt.setMessage(QByteArray::fromPercentEncoding(line));
+    self->prompt.setMessage(QByteArray::fromPercentEncoding(line).simplified());
 
     // Ugly hack here due to GnuPG being stuck in 2.0.4
     // and not having SETKEYINFO command yet. Description may contain "ID xxxxx".
     {
         const QString &str(self->prompt.message());
-        int id = str.lastIndexOf("ID ");
+        int id = str.lastIndexOf(", ID ");
```
pvuorela, line 86: What's the content like here actually? Wondering should the comment about "ID ..." be adjusted too to include the comma part.

dcaliste: That's a good remark. I've updated the comment with an example of description that can be parsed to extract the key ID.

pvuorela: Suppose the change here fixes what is says, but for another thing if the content is like above, why is this using lastIndexOf() instead of indexOf()? Quick google on such strings shows instances with email address in the end which theoretically at least might involve the id?

If fixes an issue this could be fine already.

dcaliste: The problem I got was exactly this one: with the localisation of the description by GnuPG, I ended up parsing something like `clé secrète pour l'utilisateur: « Damien Caliste <email> »
clé de 2048 bits RSA, ID A099539B, créée le 2018-04-13 (ID clé principale 41123AA4)`. Which gives, with the code in master branch, an ID like "clé prin". And I see this string in the UI, wondering what happened.

Indeed, my first intention in using `lastIndexOf()` was to avoid a possible match of "ID " in the first line containing the name and the email. But then, I didn't though about subkeys... Thus the change to parse the comma with it.

I agree this is very fragile and ugly. GnuPG in a later version is providing the key ID by itself and we don't need to parse a changing string. But this later version is GPLv3, you know.

dcaliste:
I've added the first line of the description in the comment, so it may justify the use of `lastIndexOf()`.

pvuorela (approved):
Ok, fair enough. Thanks.

PR #191, opened 2024-04-29
Fix build against OpenSSL 3 and against newer Qt

Thaodan:


lib/Crypto/Crypto.pro
```diff
@@ -4,6 +4,10 @@ TARGET = $$qtLibraryTarget($$TARGET)
 target.path = $$[QT_INSTALL_LIBS]
 CONFIG += qt create_pc create_prl no_install_prl hide_symbols
 DEFINES += SAILFISH_CRYPTO_LIBRARY_BUILD
+# QVariant::operator<` has been deprecated in Qt 5.15 however without
+# providing a replacement until Qt 6 with `QVariant::compare()`.
+# FIXME: Qt6
+QMAKE_CXXFLAGS+= -Wno-error=deprecated-declarations
```
pvuorela, line 10: This is already coming from include ../../common.pri below?

pvuorela (approved):

PR #192, opened 2024-04-29
[sailfish-secrets] Declare proper copy constructors for KeyPairGenerationParameters. JB#55609

pvuorela:
Newer compilers warn about implicitly declared copy-contstructors. Here there were almost proper explicitly defined ones but they take the parent class parameter. Such a code is also needed as KeyPairGenerationParameters::isValid() tries to create copies with parent class reference.


Thaodan (approved):
LGTM

Thaodan (approved):
LGTM

PR #197, opened 2026-02-23
[sailfish-secrets] Add a method to obtain an identifier

nephros:
which appears to be a missing piece to be able to use Secrets from QML


nephros:
I might be wrong here, but IME it's currenly not possible to use e.g. an QML
StoredSecretRequest or DeleteSecretRequest without going through c++, as
identifiers can not be created in QML.

Which basically renders large parts of Secrets useless when using QML only.

